Issue #94, opened 2022-05-02
Generate `deb` package on release


fujiapple852:
Generate `deb` packages on release using [`cargo-deb`](https://github.com/kornelski/cargo-deb) or similar

Marwanmhks:
https://github.com/fujiapple852/trippy/pull/324

PR #321, opened 2022-10-12
 Deb Packaging #94

Marwanmhks:


fujiapple852:
Thank you for the PR @Marwanmhks !

It would be great if we could automate the creation of the `deb` packages for each release, using something like [`cargo-deb`](https://github.com/kornelski/cargo-deb), is this something you would be able to help with?  I've updated #94 to clarify on the requirement.

Marwanmhks:
Yea ok I will check this out

Marwanmhks:
https://github.com/fujiapple852/trippy/pull/324

PR #463, opened 2023-03-26
Ensure administrator privileges on Windows

### LINKED ISSUE #451: Check for Administrator access on Windows
Trippy requires Administrator access on Windows (for raw sockets) but, unlike the other platforms, does not currently validate that the user has the required access on startup.

fujiapple852:
Closes #451

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -878,6 +878,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "paste"
+version = "1.0.12"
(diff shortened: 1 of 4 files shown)


zarkdav:
Should not the paste dependency be Windows specific in cargo.toml?

fujiapple852:
> Should not the paste dependency be Windows specific in cargo.toml?

Agreed, changed.

PR #537, opened 2023-05-02
feat!: use XDG conventions on macOS too

### LINKED ISSUE #540: Use XDG config dir on macOS

### LINKED ISSUE #541: Divest `dirs` crate
Divert `dirs` crate due to licensing change

utkarshgupta137:
`~/Library/Application Support` is an utterly inconvenient location for macOS users. Most CLI tools follow the `XDG conventions` on macOS as well. Replace the `dirs` crate with the `etcetera` crate to support it on macOS as well. Config locations for other platforms are unaffected.

Closes #539.
Closes #540.
Closes #541.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1272,10 +1272,11 @@ pub mod config_file {
         MultipathStrategyConfig, Protocol, TuiColor, TuiKeyBinding,
     };
     use anyhow::Context;
+    use etcetera::BaseStrategy;
     use serde::Deserialize;
     use std::fs::File;
     use std::io::read_to_string;
-    use std::path::{Path, PathBuf};
+    use std::path::Path;
 
     const DEFAULT_CONFIG_FILE: &str = "trippy.toml";
     const DEFAULT_HIDDEN_CONFIG_FILE: &str = ".trippy.toml";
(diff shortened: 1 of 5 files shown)


fujiapple852:
Thank you for the PR @utkarshgupta137!

I'm a macOS user and agree that `~/Library/Application Support` is not somewhere I'd typically place config files for command line tools.  I'd say we can drop the macOS specific dir entirely and use only XDG for nix.

Given an alternative to the `dirs` crate is needed anyway due to the new licensing, and given we need something more than the `home` crate, I think `etcetera` may be a suitable alternative.

I've reworked the code slight, I've pushed a fixup commit in https://github.com/fujiapple852/trippy/pull/539 with my suggested changes, if you're happy with them we can squash the commits and merge it.

Thanks!

src/config.rs
```diff
@@ -1285,41 +1286,39 @@ pub mod config_file {
...
-            Ok(Some(file))
-        } else if let Some(file) = read_file(dirs::home_dir, DEFAULT_HIDDEN_CONFIG_FILE)? {
-            Ok(Some(file))
-        } else if let Some(file) = read_file(dirs::config_dir, DEFAULT_CONFIG_FILE)? {
-            Ok(Some(file))
-        } else if let Some(file) = read_file(dirs::config_dir, DEFAULT_HIDDEN_CONFIG_FILE)? {
-            Ok(Some(file))
-        } else {
-            Ok(None)
+        if let Some(file) = read_file("")? {
```
fujiapple852, line 1299: I've reworked this in my commit to avoid the early returns

src/config.rs
```diff
@@ -1285,41 +1286,39 @@ pub mod config_file {
...
+        if let Some(file) = read_file(home_dir)? {
+            return Ok(Some(file));
+        }
+
+        let config_dir = etcetera::base_strategy::choose_base_strategy()?.config_dir();
+        if let Some(file) = read_file(config_dir)? {
+            return Ok(Some(file));
         }
+
+        #[cfg(target_os = "macos")]
```
fujiapple852, line 1313: Not a fan of platform specific cfg in the application layer.  We could just check for the macOS dir on all platforms but in the end I think we can just ignore the official macOS config dir and stick to XDG.

src/config.rs
```diff
@@ -1329,20 +1328,14 @@ pub mod config_file {
         Ok(toml::from_str(&read_to_string(file)?)?)
     }
 
-    fn read_file<F: FnOnce() -> Option<PathBuf>>(
-        dir: F,
-        file: &str,
-    ) -> anyhow::Result<Option<ConfigFile>> {
-        if let Some(mut path) = dir() {
-            path.push(file);
+    fn read_file<P: AsRef<Path>>(dir: P) -> anyhow::Result<Option<ConfigFile>> {
```
fujiapple852, line 1331: I added a `read_files` function to check for both the non-hidden and hidden files to keep `read_file` simple.

src/config.rs
utkarshgupta137: Yeah, I don't think anyone really wants to use the Library path. I only included it because I thought you might not want a breaking change.

utkarshgupta137:
@fujiapple852 I've included your changes & made a small modification to call `choose_base_strategy` only once. It calls `home::home_dir`, so it is better to not call it twice.

src/config.rs
fujiapple852: Good thinking, luckily the config file is new in `0.8.0` which is unreleased so there is no breaking change here.

fujiapple852:
LGTM

utkarshgupta137:
Good to merge

PR #705, opened 2023-10-03
Ipv4 and Ipv6 packet truncation

### LINKED ISSUE #686: Bug: Ipv4Packet::get_total_length() incorrect byte order
The Ipv4 `total_length` has a platform dependant byte order (see `byte_order.rs` for details) and therefore this function does not always return the correct byte order:

https://github.com/fujiapple852/trippy/blob/ad1aff226d02215398b4d78695a631e5bb91ba91/src/tracing/packet/ipv4.rs#L75-L77

This in turn can lead to the payload length calculation being incorrect:

https://github.com/fujiapple852/trippy/blob/ad1aff226d02215398b4d78695a631e5bb91ba91/src/tracing/packet/ipv4.rs#L216-L218

This is not causing a real world issue today as, when the byte order is interpreted incorrectly it typically leads to a very large value for `total_length` but this is harmless as the code takes the minimum of this and the total buffer size.

To fix this we would require access to the calculated `PlatformIpv4FieldByteOrder` in the latter function, however this is not something we wish to expose to the raw packet modules.  Instead we defer the issue to the caller by returning all available bytes after the Ipv4 header and options.  This behaviour is also consistent with the behaviour of all other `payload` methods in the packet modules.

fujiapple852:
Closes #686 

Todo:

- [x] Add Windows support for returning `bytes_read`
- [ ] Test on all platforms (different platforms may or may not adjust the Ipv4 total_length for raw sockets)

### DIFF
--- a/src/tracing/net/ipv4.rs
+++ b/src/tracing/net/ipv4.rs
@@ -190,8 +190,8 @@ pub fn recv_icmp_probe(
 ) -> TraceResult<Option<ProbeResponse>> {
     let mut buf = [0_u8; MAX_PACKET_SIZE];
     match recv_socket.read(&mut buf) {
-        Ok(_bytes_read) => {
-            let ipv4 = Ipv4Packet::new_view(&buf).req()?;
+        Ok(bytes_read) => {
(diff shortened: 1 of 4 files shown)


fujiapple852:
See #715 and #716

PR #706, opened 2023-10-04
feat(net): return the actual number of bytes read in `TracerSocket::read` and `TracerSocket::recv_from` calls on Windows

fujiapple852:


zarkdav (approved):
Look sound to me. Tested on a Windows machine and did not detect any issue.

PR #748, opened 2023-11-01
Add `dns_resolve_all` flag to trace to all resolved ips (#743)

### LINKED ISSUE #743: Trace to all IPs for a dns entry
Add an option such that if the user run `trip example.com` and `example.com` resolves to multiple IPv4/IPv6 address then, for ICMP mode, trace to all addresses simultaneously (on per tab)

fujiapple852:
Closes #743


fujiapple852:
I'm not sold on the name of the flag, `dns_resolve_all`, ideas welcome.

c-git:
Can't think of a better name

PR #788, opened 2023-11-13
Add icmp argument to allow shortcut for base protocol

### LINKED ISSUE #649: Add `--icmp` command line shortcut
Users may specify `--udp` and `--tcp` as shortcuts to `--protocol udp` (`-p udp`) and `--protocol tcp` (`-p tcp`) respectively.  

For example for `udp`:

```rust
/// Trace using the UDP protocol
#[arg(
    long,
    display_order = 3,
    conflicts_with = "protocol",
    conflicts_with = "tcp"
)]
pub udp: bool,
```

If neither is specified then the `icmp` protocol is chosen by default:

 ```rust
/// The default value for `protocol`.
pub const DEFAULT_STRATEGY_PROTOCOL: Protocol = Protocol::Icmp;
```

```rust
let protocol = cfg_layer(
    args.protocol,
    cfg_file_strategy.protocol,
    constants::DEFAULT_STRATEGY_PROTOCOL,
);
```

However the protocol may also be specified in the config file:

```
#
# Tracing strategy configuration.
#
[strategy]

# The tracing protocol.
#
# Allowed values are:
#   icmp [default]
#   udp
#   tcp
protocol = "icmp"
```

The user may wish to override the value from the config file with another protocol specified via the command line.  To do so for `icmp` the user must use the longer `--protocol icmp` (`-p icmp`) form as there is no `--icmp` shortcut.  Therefore it would be useful to add a `--icmp` command line shortcut.

trkelly23:
The following tests for the mentioned use case:
1) Change protocol in the config file to non default tcp or udp.
2) Run trippy with shortcut --icmp
Results:
Config: protocol=icmp(v4, unprivileged) as-info=n/a details=off max-hosts=auto 

This passed the basic use case please post comments if others should be run

Closes #649

### DIFF
--- a/src/config/cmd.rs
+++ b/src/config/cmd.rs
@@ -34,12 +34,31 @@ pub struct Args {
     pub protocol: Option<Protocol>,
 
     /// Trace using the UDP protocol
-    #[arg(long, conflicts_with = "protocol", conflicts_with = "tcp")]
+    #[arg(
+        long,
+        conflicts_with = "protocol",
+        conflicts_with = "tcp",
+        conflicts_with = "icmp"
+    )]
     pub udp: bool,
 
     /// Trace using the TCP protocol
-    #[arg(long, conflicts_with = "protocol", conflicts_with = "udp")]
+    #[arg(
+        long,
+        conflicts_with = "protocol",
+        conflicts_with = "udp",
+        conflicts_with = "icmp"
+    )]
     pub tcp: bool,
+
+    /// Trace using the ICMP protocol
+    #[arg(
+        long,
+        conflicts_with = "protocol",
+        conflicts_with = "udp",
+        conflicts_with = "tcp"
+    )]
+    pub icmp: bool,
 
     /// Use IPv4 only
     #[arg(short = '4', long, conflicts_with = "ipv6")]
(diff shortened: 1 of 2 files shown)


src/config/cmd.rs
```diff
@@ -41,6 +41,10 @@ pub struct Args {
     #[arg(long, conflicts_with = "protocol", conflicts_with = "udp")]
```
fujiapple852, line 41: Should also conflict with `icmp` I guess?

fujiapple852: Same for `udp` above, should conflict with `icmp` as well.

fujiapple852:
Thanks @trkelly23 - LGTM aside from small points above and the formatting (just run a default `cargo fmt` to fix that)

src/config/cmd.rs
trkelly23: Good catch - I added in the missing conflicts.

trkelly23:
PR is squashed and the changes are ready to merge.

fujiapple852:
Thanks, could you change the commit message as well? It’s a bit unclear as-is. My suggestion would be

feat: add icmp command shortcut (#649)

trkelly23:
Oops I missed the commit message which is all set and ready to merge.

fujiapple852:
Merged, thanks @trkelly23 !

PR #837, opened 2023-11-28
fix(net): crash on windows - WIP

fujiapple852:


fujiapple852:
Works, but #838 is a better fix

PR #841, opened 2023-11-29
refactor(net): use exclusive `&mut` for socket operations

### LINKED ISSUE #843: Use exclusive reference `&mut` for all Socket operations
The Socket abstracts over an OS socket and is not thread safe.  Many Socket operations mutate internal buffers.  By using exclusive (&mut) references in all Socket operations we solve both of these issues.

fujiapple852:
Closes #843


fujiapple852:
Observed error when testing on Windows (`icmp`, with 250ms rounds) for `SendTo` call:

`operation on a socket could not be performed because the system lacked sufficient buffer space or because a queue was full`

Unclear if related to this change.

fujiapple852:
Superseded by #868

PR #863, opened 2023-12-13
feat: Allow Custom Columns TUI

trkelly23:
Added:
Custom columns to TUI interface by either the configuration file or command line option
Calculation of column percentage with 40% constant percentage for hosts column.
Unit tests for TuiColumns and Columns classes 
cargo test frontend::columns
cargo test config::columns

Tests run:
cargo run -- -u --tui-custom-columns="" -m tui freebsd.org
cargo run -- -u -c ./trippy-config-sample.toml -m tui freebsd.org
cargo run -- -u --tui-custom-columns="" -c ./trippy-config-sample.toml -m tui freebsd.org
cargo run -- -u --tui-custom-columns="" -m tui freebsd.org
cargo run -- -u --tui-custom-columns="thobwd" -m tui freebsd.org

### DIFF
--- a/src/config/columns.rs
+++ b/src/config/columns.rs
@@ -1,0 +1,207 @@
+use anyhow::anyhow;
+use itertools::Itertools;
+use std::collections::HashSet;
+use std::fmt::{Display, Formatter};
+
+/// The columns to display in the hops table of the TUI.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct TuiColumns(pub Vec<TuiColumn>);
+
+impl TryFrom<&str> for TuiColumns {
+    type Error = anyhow::Error;
+
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
+        Ok(Self(
+            value
+                .chars()
+                .map(TuiColumn::try_from)
+                .collect::<Result<Vec<_>, Self::Error>>()?,
+        ))
+    }
+}
+
+impl Default for TuiColumns {
+    fn default() -> Self {
+        Self::try_from(super::constants::DEFAULT_CUSTOM_COLUMNS).expect("custom columns")
+    }
+}
+
+impl TuiColumns {
+    /// Validate the columns.
+    ///
+    /// Returns any duplicate columns.
+    pub fn find_duplicates(&self) -> Vec<String> {
+        let (_, duplicates) = self.0.iter().fold(
+            (HashSet::<TuiColumn>::new(), Vec::new()),
+            |(mut all, mut dups), column| {
+                if all.iter().contains(column) {
+                    dups.push(column.to_string());
+                } else {
+                    all.insert(*column);
+                }
+                (all, dups)
+            },
+        );
+        duplicates
+    }
+}
+
+/// A TUI hops table column.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+pub enum TuiColumn {
+    /// The ttl for a hop.
+    Ttl,
+    /// The hostname for a hostname.
+    Host,
+    /// The packet loss % for a hop.
+    LossPct,
+    /// The number of probes sent for a hop.
+    Sent,
+    /// The number of responses received for a hop.
+    Received,
+    /// The last RTT for a hop.
+    Last,
+    /// The rolling average RTT for a hop.
+    Average,
+    /// The best RTT for a hop.
+    Best,
+    /// The worst RTT for a hop.
+    Worst,
+    /// The stddev of RTT for a hop.
+    StdDev,
+    /// The status of a hop.
+    Status,
+}
+
+impl TryFrom<char> for TuiColumn {
+    type Error = anyhow::Error;
+
+    fn try_from(value: char) -> Result<Self, Self::Error> {
+        match value.to_ascii_lowercase() {
+            'h' => Ok(Self::Ttl),
+            'o' => Ok(Self::Host),
+            'l' => Ok(Self::LossPct),
+            's' => Ok(Self::Sent),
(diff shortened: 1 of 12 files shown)


src/frontend/render/settings.rs
```diff
@@ -15,7 +15,7 @@ use trippy::tracing::PortDirection;
 pub fn render(f: &mut Frame<'_>, app: &mut TuiApp) {
     let all_settings = format_all_settings(app);
     let (name, info, items) = &all_settings[app.settings_tab_selected];
-    let area = util::centered_rect(60, 60, f.size());
+    let area = util::centered_rect(75, 60, f.size());
```
fujiapple852, line 18: Why do we need to do this?

src/frontend/render/settings.rs
```diff
@@ -78,7 +78,7 @@ fn render_settings_table(
         .max()
         .unwrap_or_default()
         .max(30);
-    let table_widths = [Constraint::Min(item_width), Constraint::Length(60)];
+    let table_widths = [Constraint::Min(item_width), Constraint::Length(80)];
```
fujiapple852, line 81: As above

src/frontend/render/settings.rs
```diff
@@ -177,6 +177,10 @@ fn format_tui_settings(app: &TuiApp) -> Vec<SettingsItem> {
                 .max_addrs
                 .map_or_else(|| String::from("auto"), |m| m.to_string()),
         ),
+        SettingsItem::new(
+            "tui-custom-columns",
+            format!("{:?}", app.tui_config.tui_columns),
```
fujiapple852, line 182: Currently renders as:

```
Columns([Ttl, Host, LossPct, Sent, Received, Last, Average, Best, Worst, StdDev,
```

(gets truncated in the TUI)

Maybe just show in the `HOLSRAVBWDT` form?

src/frontend/columns.rs
```diff
@@ -0,0 +1,159 @@
...
+            | Self::Average
+            | Self::Best
+            | Self::Worst
+            | Self::StdDev
+            | Self::Ttl
+            | Self::Status => 60 / adjusted_count,
+        }
+    }
+}
+mod tests {
```
fujiapple852, line 97: Missing: `#[cfg(test)]`

src/frontend/render/table.rs
```diff
@@ -118,7 +120,38 @@ fn render_table_row(
         .bottom_margin(0)
         .style(Style::default().fg(row_color))
 }
-
+///Returns a Cell matched on short char of the Column
```
fujiapple852, line 123: Nit: 1 blank line between functions and 1 pace at the start of a comment

src/config/columns.rs
```diff
@@ -0,0 +1,198 @@
...
+            Self::Status => write!(f, "t"),
+        }
+    }
+}
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_try_from_char_for_tui_column() {
```
fujiapple852, line 119: Great tests! You could use the `#[test_case(..)]` macro here and move the negative test to a separate test.

src/frontend/columns.rs
```diff
@@ -0,0 +1,159 @@
...
+            Self::Host => 40,
+            Self::LossPct
+            | Self::Sent
+            | Self::Received
+            | Self::Last
+            | Self::Average
+            | Self::Best
+            | Self::Worst
+            | Self::StdDev
+            | Self::Ttl
```
fujiapple852, line 92: The `Ttl` (`#`) column was 3 before and `Host` was 38

src/frontend/render/settings.rs
trkelly23: Ahh a leftover from the previous output of ['H', 'O', 'L', 'S', 'R', 'A', 'V', 'B', 'W', 'D', 'T'] which got truncated in the setting pop-up.  The suggestion of displaying HOLSRAVBWDT should fit in the 60,60 rect. 
Plan is to revert to 60,60 rect.

src/frontend/render/settings.rs
trkelly23: Same as above.

src/frontend/render/settings.rs
trkelly23: As mentioned above I will be changing to display HOLSRAVBWDT.

src/frontend/columns.rs
trkelly23: Noted will tidy this up.

src/frontend/render/table.rs
trkelly23: Noted

src/config/columns.rs
trkelly23: I like the test_case macro much cleaner and easier to identify the failures.

src/frontend/columns.rs
trkelly23: I agreed with where we landed on in our conversation which was to hardcode the percentages.

_Maybe better if we just stick with hard coded widths per column type and let the TUI take care of layout
I was also thinking, if we want to calculate column widths, it would be better to add a method to the Columns type, as it knows what columns as defined, rather than having to pass a column count into the width_pct method
Lets just make it hardcoded widths for this PR, we can make it smarter later_

src/frontend/render/settings.rs
trkelly23: Turned out to be lowercase when displayed which might be preferred by users when setting columns.  Who would hit shift unless they really had to ;)

fujiapple852:
LGTM !

Issue #864, opened 2023-12-14
Automatically fall back to IPv6 if host has no IPv4 addresses


artizirk:
Currently if you don't specify `-6` argument then trippy fails to resolve addresses for IPv6 only hosts and exits with an error. Default seems to be set IPv4. It would be great if it could use any address family available.

artizirk:
Hi!

Flags `-4` and `-6` should still force `Ipv4Only` and `Ipv6Only` modes. Default without those arguments should be `trust-dns` `Ipv4thenIpv6` strategy.

I have created an PR #872 that implements this.

fujiapple852:
Hi again @artizirk, thanks for the PR!

Thinking about this a bit more, some considerations below.

### Defaults and overrides

> Default without those arguments should be trust-dns `Ipv4thenIpv6` strategy

I was a bit uneasy about this at first, as it seems to express an option of favouring Ipv4 over Ipv6.  However...

Trippy currently supports both the `addr-family` config file entry and the `--ipv4` (`-4`) and `--ipv6` (`-6`) command line flags, all of which are optional.  The command line flags (which are mutually exclusive) take precedence over the config file entry and if neither are provided there is a default of Ipv4.  

Therefore Trippy is _already_ opinionated as it defaults to Ipv4.  

However, users are able to _change_ the default opinion in the config file (`addr-family`), and so we would need to provide the same capability here, by exposing all modes (`Ipv4Only`, `Ipv6Only`, `Ipv4thenIpv6` & `Ipv6thenIpv4`) in the config file for `addr-family` as well.

> Flags -4 and -6 should still force Ipv4Only and Ipv6Only modes.

Yes I think that is right.  The `--ipv4` (`-4`) and `--ipv6` (`-6`) can stay as shortcuts for `Ipv4Only` and `Ipv6Only` respectively, and continue to override whatever is set in the config file.

### System Resolver

The logic for the `system` resolver should match the logic for the other resolvers (`google`, `cloudflare` etc) which use `trust-dns`, which is an implementation detail users should not be aware of (fyi as great as it is, I'm looking to swap it for an non-async DNS resolver crate if I can).  This means we need to find a way to emulate the various modes provided by `trust-dns` in the `system` resolver.  

[...]

PR #872, opened 2023-12-15
feat: default resolve both IPv4 and IPv6 addresses

artizirk:
This is useful when target host only has a IPv6 address and the user has not forced particular address family.

Fixes: #864


artizirk:
NB: I have never written any rust before this so I may have missed something.

fujiapple852:
Thanks for the PR @artizirk!  I've built on the idea from this PR in #928 which supersedes it and so I will close this.

PR #881, opened 2023-12-21
!feat(cfg): replace `mode` (`-m`) flag with sub-commands (#880) - WIP

### LINKED ISSUE #880: [BREAKING CHANGE] Replace `mode` with sub-commands
Replace the `mode` (`-m`) flag with sub-commands to allow specifying sub-command specific command line flags.

fujiapple852:
Closes #880

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -25,6 +25,7 @@ mod constants;
 mod file;
 mod theme;
 
+use crate::config::cmd::{Commands, CommonArgs, DotArgs, FlowsArgs, ReportArgs, TuiArgs};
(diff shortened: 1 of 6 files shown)


fujiapple852:
Issues:

Doesn't check for positional arguments (target):

```shell
$ ./target/debug/trip -u
Error: failed to find any valid IPv4 addresses for
```

Should be mandatory specify the report type:

```shell
$ ./target/debug/trip report
Error: privileges are required (hint: try adding -u to run in unprivileged mode)

see https://github.com/fujiapple852/trippy#privileges for details
```

Should we have reports at all or just have sub-commands for `pretty`, `csv` etc?

PR #882, opened 2023-12-22
feat display jitter

### LINKED ISSUE #39: Calculate and display Jitter
Trippy current calculates the `StDev` column as:

```rust
pub fn stddev_ms(&self) -> f64 {
    if self.total_recv > 1 {
        (self.m2 / (self.total_recv - 1) as f64).sqrt()
    } else {
        0_f64
    }
}
```

Where, for each successful `Probe` (for a given hop `ttl`), the `total_recv`, `mean` and `m2` values are updated as follows:

```rust
let dur = probe.duration();
let dur_ms = dur.as_secs_f64() * 1000_f64;
hop.total_recv += 1;
hop.mean += (dur_ms - hop.mean) / hop.total_recv as f64;
hop.m2 += (dur_ms - hop.mean) * (dur_ms - hop.mean);
```

We would like to add support for a Jitter columns similar to those provided by mtr (see https://networkengineering.stackexchange.com/questions/5288/what-is-interarrival-jitter-in-mtr)

trkelly23:
The branch is currently a WIP to get feedback on Jitter calculations and display.

Know issues:
The display will overflow if all the columns are displayed; though as an interim fix, the Constraint::Min is used.  
Jitter columns will only show using tui-custom-columns 
Jitter columns display in json & tui mode's only.  Json was added for saving test results for comparison to mtr.

Closes #39

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,6 +52,7 @@ tracing-subscriber = { version = "0.3.18", default-features = false, features = 
 tracing-chrome = "0.7.1"
 petgraph = "0.6.4"
 csv = "1.3.0"
+iso8601-timestamp = "0.2.16"
 serde_with = "3.4.0"
 
 # Library dependencies (Linux)
--- a/src/backend/trace.rs
+++ b/src/backend/trace.rs
@@ -144,6 +144,14 @@ pub struct Hop {
     extensions: Option<Extensions>,
     mean: f64,
     m2: f64,
+    /// The ABS(RTTx - RTTx-n)
+    jitter: Option<Duration>,
+    /// The Sequential jitter average calculated for each
+    javg: Option<f64>,
+    /// The worst jitter reading recorded
+    jmax: Option<Duration>,
+    /// The interval calculation i.e smooth
+    jinta: f64,
 }
 
 impl Hop {
@@ -219,6 +227,31 @@ impl Hop {
         }
     }
 
+    /// The duration of the jitter probe observed.
+    pub fn jitter_ms(&self) -> Option<f64> {
+        self.jitter.map(|j| j.as_secs_f64() * 1000_f64)
+    }
+    /// The duration of the jworst probe observed.
+    pub fn jmax_ms(&self) -> Option<f64> {
+        self.jmax.map(|x| x.as_secs_f64() * 1000_f64)
+    }
+    /// The jitter average duration of all probes.
+    pub fn javg_ms(&self) -> Option<f64> {
+        if self.total_recv() > 0 {
+            self.javg
+        } else {
+            None
+        }
+    }
+    /// The jitter interval of all probes.
+    pub fn jinta(&self) -> Option<f64> {
+        if self.total_recv() > 0 {
+            Some(self.jinta)
+        } else {
+            None
+        }
+    }
+
     /// The last N samples.
     pub fn samples(&self) -> &[Duration] {
         &self.samples
@@ -244,6 +277,10 @@ impl Default for Hop {
             m2: 0f64,
             samples: Vec::default(),
             extensions: None,
+            jitter: None,
+            javg: None,
+            jmax: None,
+            jinta: 0f64,
         }
     }
 }
@@ -336,6 +373,21 @@ impl TraceData {
                 let dur = probe.duration();
                 let dur_ms = dur.as_secs_f64() * 1000_f64;
                 hop.total_time += dur;
+                //Before last is set use it to calc jitter
+                let last_ms = hop.last_ms().unwrap_or_default();
+                let jitter_ms = (last_ms - dur_ms).abs();
+                let jitter_dur = Duration::from_secs_f64(jitter_ms / 1000_f64);
+                hop.jitter = Some(jitter_dur);
+                let mut javg_ms = hop.javg_ms().unwrap_or_default();
+                //Welfords online algorithm avg without dataset values.
+                javg_ms += (jitter_ms - javg_ms) / hop.total_recv as f64;
+                hop.javg = Some(javg_ms);
+                // algorithm is from rfc1889, A.8 or rfc3550
+                hop.jinta += jitter_ms - ((hop.jinta + 8.0) / 16.0);
+                //Max following logic of hop.worst
+                hop.jmax = hop
+                    .jmax
+                    .map_or(Some(jitter_dur), |d| Some(d.max(jitter_dur)));
                 hop.last = Some(dur);
                 hop.samples.insert(0, dur);
                 hop.best = hop.best.map_or(Some(dur), |d| Some(d.min(dur)));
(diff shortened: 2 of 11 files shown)


src/backend/trace.rs
```diff
@@ -336,6 +374,18 @@ impl TraceData {
                 let dur = probe.duration();
                 let dur_ms = dur.as_secs_f64() * 1000_f64;
                 hop.total_time += dur;
+                //Before last is set use it to set jitter
+                let last_ms = hop.last_ms().map_or(0f64, |d|d);
```
fujiapple852, line 378: You could use `unwrap_or_default()` here

src/backend/trace.rs
```diff
@@ -336,6 +374,18 @@ impl TraceData {
...
+                let jitter_ms = (last_ms - dur_ms).abs();
+                let jitter_dur = Duration::from_millis(jitter_ms as u64);
+                hop.jitter = Some(jitter_dur);
```
fujiapple852, line 381: This is set to be `Some` unconditionally, I guess we should only set this if `hop.last` is `Some`?

src/backend/trace.rs
```diff
@@ -144,6 +144,15 @@ pub struct Hop {
     extensions: Option<Extensions>,
     mean: f64,
     m2: f64,
+    /// The ABS(RTTx - RTTx-n)
+    jitter: Option<Duration>,
+    /// The Sequential jitter average calculated for each
+    javg: f64,
```
fujiapple852, line 150: Given the logic in fn `javg_ms` to return this as an `Option`, perhaps it would make sense for `javg` to be an `Option`?

src/backend/trace.rs
```diff
@@ -336,6 +374,18 @@ impl TraceData {
                 let dur = probe.duration();
                 let dur_ms = dur.as_secs_f64() * 1000_f64;
                 hop.total_time += dur;
+                //Before last is set use it to set jitter
+                let last_ms = hop.last_ms().map_or(0f64, |d|d);
+                let jitter_ms = (last_ms - dur_ms).abs();
+                let jitter_dur = Duration::from_millis(jitter_ms as u64);
```
fujiapple852, line 380: I wonder if we can achieve this calculation without the round trip from `Duration` to `f64` via `u64`?

src/backend/trace.rs
```diff
@@ -336,6 +374,18 @@ impl TraceData {
...
+                hop.jitter = Some(jitter_dur);
+                let oldjavg = hop.javg;
+                //Welfords online alogrithm to get avg without all values.
+                hop.javg += (jitter_ms - oldjavg) / hop.total_recv as f64;
```
fujiapple852, line 384: should be safe to do this and avoid needing `oldjavg`:

```
 hop.javg += (jitter_ms - hop.javg) / hop.total_recv as f64;
```

fujiapple852:
@trkelly23 I've added a few minor review comments on the impl but aside fro that the logic of all four calculations (`jttr`, `javg`, `jmax`, `jinta`) look correct to me.  Have you been able to run a side by side comparison with mtr eyeball the results?

fujiapple852:
> Display will overflow if all the columns are displayed.

Yes, I think we're now going to have to solve the general issue of table column sizing when we have a variable number of columns with varying minimum size requirements.  The switch here to use `Min` has a fairly dramatic effect on the table layout and i'm not convinced that is what we want here.  What do you think the right solution is here?  I'll have a think as well.

> Jitter columns will only show using tui-custom-columns

I think this is fine for now.  There are lots of other columns I would like to add that would not be shown be default (for example, estimate geographic distance between hops when both GeoIps are known).  As we discussed before, in the future we may want to add a dialog for dynamic column selection.

trkelly23:
> @trkelly23 I've added a few minor review comments on the impl but aside from that the logic of all four calculations (`jttr`, `javg`, `jmax`, `jinta`) looks correct to me. Have you been able to run a side by side comparison with mtr to eyeball the results?

I have run a many side by side tests the best comparison method was to add the jitter fields to trippy JSON output and run both.
image
Above are typical results some values are very close or not.  Has this been your experience?
Any suggestions on other tests?

A set of results run 1/1/2024.
[1_trip_test.json](https://github.com/fujiapple852/trippy/files/13804987/1_trip_test.json)
[2_trip_test.json](https://github.com/fujiapple852/trippy/files/13804988/2_trip_test.json)
[2_mtr_test.json](https://github.com/fujiapple852/trippy/files/13804989/2_mtr_test.json)
[1_mtr_test.json](https://github.com/fujiapple852/trippy/files/13804990/1_mtr_test.json)

src/backend/trace.rs
trkelly23: I went ahead and made it an option.   I think this is an improvement and keeps the operations consistent for the Hop members.

src/backend/trace.rs
trkelly23: let jitter_dur = Duration::from_secs_f64(jitter_ms / 1000_f64);
It's now dividing the millis by 1000 i.e. moving the decimal place.  I see this being done on other durations and should be safe.

src/backend/trace.rs
trkelly23: let mut javg_ms = hop.javg_ms().unwrap_or_default();
                //Welfords online algorithm avg without dataset values.
                javg_ms += (jitter_ms - javg_ms) / hop.total_recv as f64;
                hop.javg = Some(javg_ms);

Agreed it is safe however a little extra work is required because it's a cumulative average(+=). See above.

src/backend/trace.rs
```diff
@@ -381,3 +433,240 @@ impl TraceData {
...
+                worst: Some(Duration::from_millis(worst)),
+                mean,
+                m2,
+                jitter: Some(Duration::from_millis(jitter)),
+                javg: Some(javg),
+                jmax: Some(Duration::from_millis(jmax)),
+                jinta,
+            }
+        }
+        //TODO: Create combined struct ImportProbe Vec<ImportProbe> & HopResults.
```
fujiapple852, line 578: I think that is a good idea.  Can we use yaml rather than json as well?

Cargo.toml
```diff
@@ -52,6 +52,7 @@ tracing-subscriber = { version = "0.3.18", default-features = false, features =
 tracing-chrome = "0.7.1"
 petgraph = "0.6.4"
 csv = "1.3.0"
+iso8601-timestamp = "0.2.16"
```
fujiapple852, line 55: Can we avoid needing this dep?  Can `humantime` be used instead?  Note it could be in `[dev-dependancies]` as well.

src/backend/trace.rs
```diff
@@ -381,3 +433,240 @@ impl TraceData {
...
+            && self.jitter == other.jitter
+            && self.jmax == other.jmax
+            && self.javg == other.javg
+            && format!("{:.2}",self.jinta) == format!("{:.2}", other.jinta)
+            }
+    }    
+    #[test_case("base_line.json", &HopResults::new(2,2,700,300,700,0.0,0.0,400,350.0,400,680.28))]
+    fn test_probe_file(json_file: &str, expected: &HopResults) {
+        let mut json = String::new();
+        let mut dir_json_file = "./tests/data/".to_owned();
```
fujiapple852, line 593: Can we use include_str!() to read these at compile time?  See https://github.com/fujiapple852/trippy/blob/55641c6a60bde36a3a08f72f15b6dfa49215d994/tests/sim/tests.rs#L31 for an example

src/backend/trace.rs
trkelly23: After looking at the TUN test code I had the same exact thought, make all test input files the same type.
Consider it done!

Cargo.toml
trkelly23: I see input the difference (RTTx - RTTx-1) - I can easily move to this type of model.
Yes, it should only be needed as a [dev-dependancies].

fujiapple852:
Superseded by #937

PR #924, opened 2024-01-14
feat(tui): improve column width layout logic

### LINKED ISSUE #925: Improve table column layout constraint logic
Each column is designated as either `Fixed(n)` or `Variable` and then to compute the width of each column we use the size `n` for `Fixed` columns and compute the size of the `Variable` columns as total table width - sum(fixed size columns) divided by number number of variable size columns.

The effect is that all columns remain fixed in size except the `Host` column which is `Variable` and can therefore grows or shrinks as the window is resized or columns are added or removed.

This may be replaced in future by a more complex layout algorithm but works well enough for today.

fujiapple852:
Closes #925

### DIFF
--- a/src/frontend/columns.rs
+++ b/src/frontend/columns.rs
@@ -1,9 +1,45 @@
 use crate::config::{TuiColumn, TuiColumns};
+use ratatui::layout::{Constraint, Rect};
 use std::fmt::{Display, Formatter};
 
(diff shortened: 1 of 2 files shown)


fujiapple852:
@trkelly23 LMK what you think of this and how well it works with custom columns.

PR #955, opened 2024-01-26
test: add backend tracer tests

trkelly23:
The PR includes tests for a single hop and 0 to n probes for various probe statuses.  
Note: I did not add mean & m2 to the equality test but plan to add them later.
During tests, it was found that low latency of .5 millis the Jinta value turned negative.  According to the RFC near zero should remain 0 because is minimal.  Added check jinta.max(0.5)

A couple of items to please review:
1) I made tracer & type(src/tracing.rs) public to integrate with tests.  Not completely kosher as some libs are private and public.
2)  I'm looking for a suggestion to add multiple hops to tests.  Can you please suggest it?
3) Are there other missing test cases besides multiple hops?
4) The location of the YAML files is under the backend folder.  I ran into multiple issues moving outside the src directory.  Stumped on it.

### DIFF
--- a/src/backend/trace.rs
+++ b/src/backend/trace.rs
@@ -410,7 +410,7 @@ impl TraceData {
                 hop.total_time += dur;
                 // Before last is set use it to calc jitter
                 let last_ms = hop.last_ms().unwrap_or_default();
-                let jitter_ms = (last_ms - dur_ms).abs();
+                let jitter_ms = (dur_ms - last_ms).abs();
                 let jitter_dur = Duration::from_secs_f64(jitter_ms / 1000_f64);
                 hop.jitter = hop.last.and(Some(jitter_dur));
                 hop.javg += (jitter_ms - hop.javg) / hop.total_recv as f64;
--- a/src/backend/trace.rs
+++ b/src/backend/trace.rs
@@ -415,7 +415,7 @@ impl TraceData {
                 hop.jitter = hop.last.and(Some(jitter_dur));
                 hop.javg += (jitter_ms - hop.javg) / hop.total_recv as f64;
                 // algorithm is from rfc1889, A.8 or rfc3550
-                hop.jinta += jitter_ms - ((hop.jinta + 8.0) / 16.0);
+                hop.jinta += jitter_ms.max(0.5) - ((hop.jinta + 8.0) / 16.0);
                 hop.jmax = hop
                     .jmax
                     .map_or(Some(jitter_dur), |d| Some(d.max(jitter_dur)));
--- a/src/backend/trace.rs
+++ b/src/backend/trace.rs
@@ -468,3 +468,224 @@ impl TraceData {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use anyhow::anyhow;
+    use serde::Deserialize;
+    use std::collections::HashSet;
+    use std::ops::Add;
+    use std::str::FromStr;
+    use std::time::SystemTime;
+    use test_case::test_case;
+    use trippy::tracing::{
+        CompletionReason, Flags, IcmpPacketType, Port, Probe, ProbeComplete, ProbeState, Sequence,
+        TimeToLive, TraceId,
+    };
+
+    /// A test scenario.
+    #[derive(Deserialize, Debug)]
+    #[serde(deny_unknown_fields)]
+    struct Scenario {
+        /// the biggest ttl expected in this scenario
+        largest_ttl: u8,
+        /// The rounds of probe tracing data in this scenario.
+        rounds: Vec<RoundData>,
+        /// The expected outcome from running this scenario.
+        expected: Expected,
+    }
+
+    /// A single round of tracing probe data.
+    #[derive(Deserialize, Debug)]
+    #[serde(deny_unknown_fields)]
+    struct RoundData {
+        /// The probes in this round.
+        probes: Vec<ProbeData>,
+    }
+
+    /// A single probe from a single round.
+    #[derive(Deserialize, Debug)]
+    #[serde(deny_unknown_fields)]
+    #[serde(try_from = "String")]
+    struct ProbeData(ProbeState);
+
+    impl TryFrom<String> for ProbeData {
+        type Error = anyhow::Error;
+
+        fn try_from(value: String) -> Result<Self, Self::Error> {
+            // format: {ttl} {status} {duration} {host} {sequence} {src_port} {dest_port}
+            let values = value.split_ascii_whitespace().collect::<Vec<_>>();
+            if values.len() == 7 {
+                let ttl = TimeToLive(u8::from_str(values[0])?);
+                let state = values[1].to_ascii_lowercase();
+                let sequence = Sequence(u16::from_str(values[4])?);
+                let src_port = Port(u16::from_str(values[5])?);
+                let dest_port = Port(u16::from_str(values[6])?);
+                let round = Round(0); // note we inject this later, see ProbeRound
+                let sent = SystemTime::now();
+                let flags = Flags::empty();
+                let state = match state.as_str() {
+                    "n" => Ok(ProbeState::NotSent),
+                    "s" => Ok(ProbeState::Skipped),
+                    "a" => Ok(ProbeState::Awaited(Probe::new(
+                        sequence,
+                        TraceId(0),
+                        src_port,
+                        dest_port,
+                        ttl,
+                        round,
+                        sent,
+                        flags,
+                    ))),
+                    "c" => {
+                        let host = IpAddr::from_str(values[3])?;
+                        let duration = Duration::from_millis(u64::from_str(values[2])?);
+                        let received = sent.add(duration);
+                        let icmp_packet_type = IcmpPacketType::NotApplicable;
+                        Ok(ProbeState::Complete(
+                            Probe::new(
+                                sequence,
+                                TraceId(0),
+                                src_port,
+                                dest_port,
+                                ttl,
+                                round,
+                                sent,
+                                flags,
+                            )
+                            .complete(
+                                host,
+                                received,
+                                icmp_packet_type,
+                                None,
+                            ),
+                        ))
+                    }
+                    _ => Err(anyhow!("unknown probe state")),
+                }?;
+                Ok(Self(state))
+            } else {
+                Err(anyhow!("failed to parse {}", value))
+            }
+        }
+    }
+
+    /// A helper struct so wwe may inject the round into the probes.
+    struct ProbeRound(ProbeData, Round);
+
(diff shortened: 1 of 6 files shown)


fujiapple852:
@trkelly23 nice start!  

I pushed some more WIP commits (I didn't squash so you can easily see my changes, we can squash later once it's ready):

- I rebased
- I moved the test_resources to the top level folder and reorganised these and other test_resources
- I simplified the code bit and did some refactoring (removed unneeded `TryFrom`, added a `ProbeStatusTest` type)

> I made tracer & type(src/tracing.rs) public to integrate with tests. Not completely kosher as some libs are private and public.

It was just missing an re-export of `Round` which I've added, no need to make the modules public (I changed it).

> I'm looking for a suggestion to add multiple hops to tests. Can you please suggest it?
> Are there other missing test cases besides multiple hops?

What we are trying to test is the final state of the `Trace` type after the completion of several rounds (i.e. several calls to `update_from_round(TracerRound)`).

Therefore we want each test "scenario" (i.e. file) to consist of:

Input: zero or more `TracerRound`, where each round contains zero or more probes (plus `largest_ttl` and `reason`)
Output: the state of the `Trace` after all `TracerRound` have been applied

At the moment the code has:

Input: zero or more `Probe` (all for the same hop/ttl, but for different rounds, which wouldn't happen in reality)
Output: the state of a single `Hop`

> The location of the YAML files is under the backend folder. I ran into multiple issues moving outside the src directory. Stumped on it.

I moved these, seemed to work ok?

fujiapple852:
> Note: I did not add mean & m2 to the equality test but plan to add them later.

I removed these as they are "private internal state" and so not something we need to explicitly test (we're "black box" testing the `Trace` type).

> During tests, it was found that low latency of .5 millis the Jinta value turned negative. According to the RFC near zero should remain 0 because is minimal. Added check jinta.max(0.5)

Always good when adding tests spots a bug!  Can I trouble you to extract the fix into a separate "fix: xxx" commit?

trkelly23:
> > Note: I did not add mean & m2 to the equality test but plan to add them later.
> 
> I removed these as they are "private internal state" and so not something we need to explicitly test (we're "black box" testing the `Trace` type).
> 
> > During tests, it was found that low latency of .5 millis the Jinta value turned negative. According to the RFC near zero should remain 0 because is minimal. Added check jinta.max(0.5)
> 
> Always good when adding tests spots a bug! Can I trouble you to extract the fix into a separate "fix: xxx" commit?

Thanks for confirming mean and m2 don't need tests.
I'll backout the fix for low latency and put it into a different branch.

trkelly23:
> Therefore we want each test "scenario" (i.e. file) to consist of:
> 
> Input: zero or more `TracerRound`, where each round contains zero or more probes (plus `largest_ttl` and `reason`) Output: the state of the `Trace` after all `TracerRound` have been applied



> @trkelly23 nice start!
> 
> I pushed some more WIP commits (I didn't squash so you can easily see my changes, we can squash later once it's ready):
> 
> * I rebased
> * I moved the test_resources to the top level folder and reorganised these and other test_resources
> * I simplified the code bit and did some refactoring (removed unneeded `TryFrom`, added a `ProbeStatusTest` type)
LOL, I had it a From trait originally.  Good to see it back as one and the clutter removed.
> 
> > I made tracer & type(src/tracing.rs) public to integrate with tests. Not completely kosher as some libs are private and public.
> 
> It was just missing an re-export of `Round` which I've added, no need to make the modules public (I changed it).
Thanks much cleaner.
> 
> > I'm looking for a suggestion to add multiple hops to tests. Can you please suggest it?
> > Are there other missing test cases besides multiple hops?
> 
> What we are trying to test is the final state of the `Trace` type after the completion of several rounds (i.e. several calls to `update_from_round(TracerRound)`).
> 
> Therefore we want each test "scenario" (i.e. file) to consist of:
> 
> Input: zero or more `TracerRound`, where each round contains zero or more probes (plus `largest_ttl` and `reason`) Output: the state of the `Trace` after all `TracerRound` have been applied
Should senario be changed to?
Scenario {
        largest_ttl: usize,
        reason: CompletionReason,
        hop: HopTest,
        probes: Vec<ProbeTest>,
}
> 
> At the moment the code has:
> 
> Input: zero or more `Probe` (all for the same hop/ttl, but for different rounds, which wouldn't happen in reality) Output: the state of a single `Hop`
> 
> > The location of the YAML files is under the backend folder. I ran into multiple issues moving outside the src directory. Stumped on it.
> 
> I moved these, seemed to work ok?
Yes, I tested the move, and works locally.

src/backend/trace.rs
```diff
@@ -410,12 +410,12 @@ impl TraceData {
                 hop.total_time += dur;
                 // Before last is set use it to calc jitter
                 let last_ms = hop.last_ms().unwrap_or_default();
-                let jitter_ms = (last_ms - dur_ms).abs();
+                let jitter_ms = (dur_ms - last_ms).abs();
```
fujiapple852, line 413: @trkelly23 not sure if we need this change, it is the same either way since we take the absolute value.  What is the rationale here?

src/backend/trace.rs
```diff
@@ -410,12 +410,12 @@ impl TraceData {
...
                 let jitter_dur = Duration::from_secs_f64(jitter_ms / 1000_f64);
                 hop.jitter = hop.last.and(Some(jitter_dur));
                 hop.javg += (jitter_ms - hop.javg) / hop.total_recv as f64;
                 // algorithm is from rfc1889, A.8 or rfc3550
-                hop.jinta += jitter_ms - ((hop.jinta + 8.0) / 16.0);
+                hop.jinta += jitter_ms.max(0.5) - ((hop.jinta + 8.0) / 16.0);
```
fujiapple852, line 418: @trkelly23 can you help me understand the math here?

src/backend/trace.rs
```diff
@@ -468,3 +468,229 @@ impl TraceData {
...
+        avg_ms: f64,
+        jitter: Option<f64>,
+        javg: f64,
+        jmax: Option<f64>,
+        jinta: f64,
+        addrs: HashMap<IpAddr, usize>,
+        samples: Option<Vec<f64>>,
+        last_src: u16,
+        last_dest: u16,
+        last_sequence: u16,
```
fujiapple852, line 623: @trkelly23 I added all remaining fields we need to test.  I also adjusted which are `Option<T>` to match the public interface.  I only updated `test_scenario.yaml` to match for now, the other scenarios will need to be updated as well if we're happy with this approach.

src/backend/trace.rs
```diff
@@ -468,3 +468,229 @@ impl TraceData {
...
+                        .iter()
+                        .map(|s| s.as_secs_f64() * 1000_f64)
+                        .collect()
+                ),
+                expected.samples
+            );
+        }
+    }
+
+    #[allow(clippy::float_cmp)]
```
fujiapple852, line 687: @trkelly23 A simple float compare should be fine here since we're comparing against a constant value from our scenario file.

src/backend/trace.rs
```diff
@@ -468,3 +468,229 @@ impl TraceData {
...
+    use trippy::tracing::{
+        CompletionReason, Flags, IcmpPacketType, Port, Probe, ProbeComplete, ProbeState, Sequence,
+        TimeToLive, TraceId,
+    };
+
+    /// A test scenario.
+    #[derive(Deserialize, Debug)]
+    #[serde(deny_unknown_fields)]
+    struct Scenario {
+        #[allow(dead_code)]
```
fujiapple852, line 491: Maybe we just get rid of this field?

src/backend/trace.rs
```diff
@@ -468,3 +468,229 @@ impl TraceData {
...
+                .map(Into::into)
+                .collect::<Vec<_>>();
+            let largest_ttl = TimeToLive(scenario.largest_ttl);
+            let tracer_round =
+                TracerRound::new(&probes, largest_ttl, CompletionReason::TargetFound);
+            trace.update_from_round(&tracer_round);
+        }
+        let actual_hops = trace.hops(Trace::default_flow_id());
+        let expected_hops = scenario.expected.hops;
+        for (actual, expected) in actual_hops.iter().zip(expected_hops) {
```
fujiapple852, line 653: Tweaked to use `zip`

src/backend/trace.rs
```diff
@@ -468,3 +468,229 @@ impl TraceData {
...
+            );
+        }
+    }
+
+    #[allow(clippy::float_cmp)]
+    fn assert_eq_optional(actual: Option<f64>, expected: Option<f64>) {
+        match (actual, expected) {
+            (Some(actual), Some(expected)) => assert_eq!(actual, expected),
+            (Some(_), None) => panic!("actual {actual:?} but not expected"),
+            (None, Some(_)) => panic!("expected {expected:?} but no actual"),
```
fujiapple852, line 692: @trkelly23 tweaked so that we fail for cases where either expected or actual is missing.  We could relax this to allow cases where we have the actual without the expected to give us a way to "ignore" the field but I tend to prefer it being strict.

src/backend/trace.rs
```diff
@@ -468,3 +468,229 @@ impl TraceData {
...
+    #[derive(Deserialize, Debug, Clone)]
+    #[serde(deny_unknown_fields)]
+    struct Expected {
+        /// The expected outcome per hop.
+        hops: Vec<HopData>,
+    }
+
+    /// The expected outcome for a single hop.
+    #[derive(Deserialize, Debug, Clone)]
+    #[serde(deny_unknown_fields)]
```
fujiapple852, line 605: Added `deny_unknown_fields` everywhere to catch typos, there were a few

src/backend/trace.rs
trkelly23: I found a slight difference in the spec'd math of RFC3550 and changed it to align the code with the RFC.

src/backend/trace.rs
trkelly23: I found during extreme low-latency testing (jitter_ms .lt .5 millis) the jinta equation returns negative values.  I added a floor of .5 millis as a fix for the bug.
From mtr the jinta equation is jinta = jitter - ((nh->jinta + 8) >> 4);
https://github.com/traviscross/mtr/blob/master/ui/net.c#L304

I update ours to divide by 16.0 which is same as >>4.  I preferred using floats to avoid having to force a conversion to float of the result using "as f64", it a little cleaner this way.

src/backend/trace.rs
fujiapple852: ok makes sense!

src/backend/trace.rs
fujiapple852: Removed

src/backend/trace.rs
fujiapple852: I've moved this to a separate `fix` commit.

PR #956, opened 2024-01-27
ci: reduce unneeded ci runs

c-git:
Trying to reduce unnecessary CI runs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,6 @@
 on:
   pull_request:
     branches: [ master ]
-  schedule:
-    - cron: '00 18 * * *'
 
 name: CI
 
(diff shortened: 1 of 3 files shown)


c-git:
Patterned the split on https://github.com/LukeMathWalker/zero-to-production/blob/main/.github/workflows/audit.yml

c-git:
I wanted to only run CI if there were code changes to avoid running it without a good reason as it would just have the same result (hopefully) if no code changes but then I don't know how to make branch protection rules that conditionally require jobs. I would remove cargo-deny from the list because that one doesn't need to run as often.

c-git:
After having thought about it I can't justify adding  the path filters because of the additional challenges it creates regarding branch protection rules so I've removed that commit and just going to go with the easy savings only.

.github/workflows/ci.yml
```diff
@@ -1,8 +1,6 @@
 on:
   pull_request:
     branches: [ master ]
-  schedule:
-    - cron: '00 18 * * *'
```
fujiapple852, line 5: We want to run the CI job daily to make sure we run against the latest `stable`, `beta` and `nightly` Rust and the latest `macos-latest` and `windows-latest`.

aside: we should probably use `ubuntu-latest` rather than `ubuntu-20.04`, worth fixing i'd say.

fujiapple852:
@c-git it's a good thought, but given the CI jobs run in parallel and the deny job takes ~20secs vs the longest job which takes ~2+mins I don't know if this buys us much?

c-git:
No it would have only made sense if we didn't need to run the others every day. But given that we are testing also testing things other than the code which wouldn't have changed then I think it doesn't make sense to separate out the security check. Will open a separate PR to change the version of Ubuntu we use.

PR #977, opened 2024-02-01
feat: add sequence number custom column (#976)

### LINKED ISSUE #976: Add sequence custom column
Add sequence number as a custom column

fujiapple852:
Closes #976

### DIFF
--- a/src/backend/trace.rs
+++ b/src/backend/trace.rs
@@ -151,6 +151,8 @@ pub struct Hop {
     last_src_port: u16,
     /// The destination port for last probe for this hop.
     last_dest_port: u16,
+    /// The sequence number for the last probe for this hop.
+    last_sequence: u16,
     /// The history of round trip times across the last N rounds.
     samples: Vec<Duration>,
     /// The ICMP extensions for this hop.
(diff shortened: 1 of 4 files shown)


trkelly23:
Very cool to see new columns.
For any new columns, I think it's a good idea to add them to the reports Hop struct src/report/types.rs.  
Then it's a quick add to ```impl<R: Resolver> From<(&backend::trace::Hop, &R)> for Hop```

Admittedly, this is not complete across all reporting outputs the new columns will at the least be in the JSON.  I have used the JSON to compare tests.

fujiapple852:
@trkelly23 I must admin that I forgot about adding these new columns to the reports, so thank you for the reminder.  

Thinking about it now, I don't think it makes much sense to add the last ports or sequence number to the reports, as these change every round yet the report only give you a snapshot after N rounds (i.e. what the last sequence number as for the final round of the report).  

I guess that's true for the existing `last` rtt value on the report as well, which is the only other non-aggregate value we currently hold.  Instead, perhaps we should include the vec of `samples` in the report?  The same would be the case for ports and sequence, we could aggregate these into a vec across all rounds and report them all at the end, though I don't think that would be particularly useful to users.

PR #983, opened 2024-02-04
Snapshot Testing

c-git:
This version of the PR uses a sub-folder inside of main then inside of there is the folder for the snapshots. It works but then we'd need another folder for tests in each file. See [PR #984](https://github.com/fujiapple852/trippy/pull/984) for more info on overall pros and cons.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -312,6 +312,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "console"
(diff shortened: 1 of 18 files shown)


c-git:
Closing this PR as I think the other approach is better since I discovered how to set the location for the snap files and they do not need to be in the src directory.

PR #984, opened 2024-02-04
Snapshot-shared-folder

c-git:
Replace the snapshots manually managed over to insta for management
using the tool

We can choose to put the snapshot files anywhere we want so where the files are is not an issue. Only issue we need to contend with is white space and what appears to be screen with issues between local and CI.


fujiapple852:
What shall we do with this PR @c-git ?

c-git:
Sure, I don't recall what we'd decided. I'll have to try to lookup the chat.

c-git:
By any chance do you recall?

c-git:
Documenting the next planned steps

- [ ] Rebase branch to onto main
- [ ] Clear out old work done on this
- [ ] Remove the current baseline files and replace them with snapshots (hopefully they will be similar as we'll also be removing the whitespace so it will look like a file rename and not a delete and add).
- [ ] Test and document update process (where should it be documented?)

c-git:
Closed by rebase, will reopen once I have a commit I want. Just trying to finalize what to do in my head. Trying to read the current code to get context.

c-git:
I can merge it at the end instead of modify the same files multiple times but for now wanted to start from closer to the target structure.

c-git:
Ok done. Didn't document how to accept the new snapshots when the time comes but I don't know where would be a good place to do that, contributors guide?

I included a string version of the input before removing white space you'll let me know if it makes sense.

crates/trippy/Cargo.toml
```diff
@@ -53,8 +53,14 @@ clap_mangen = "0.2.20"
...
 test-case.workspace = true
 serde_yaml.workspace = true
+insta = { version = "1.39.0", features = ["serde"] }
 pretty_assertions = "1.4.0"
 
+# Improve testing speed https://insta.rs/docs/quickstart/#optional-faster-runs
+[profile.dev.package]
+insta.opt-level = 3
+similar.opt-level = 3
+
```
fujiapple852, line 63: Gives:

```
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /Users/tristan/projects/trippy/crates/trippy/Cargo.toml
workspace: /Users/tristan/projects/trippy/Cargo.toml
```

I'd suggest we remove this, I doubt the speedup is significant for the small file sizes.

fujiapple852, line 63: Gives:

```
warning: profiles for the non root package will be ignored, specify profiles at the workspace root
```

I don't think we need this anyway as I doubt the speedup is significant for the small file sizes.

crates/trippy/tests/resources/snapshots/trip__config__tests__compare_snapshot@trip --help.snap
fujiapple852, line 1: Can we control the file names?  I'm a little uneasy about filenames with spaces.

crates/trippy/tests/resources/snapshots/trip__config__tests__compare_snapshot@trip -h.snap
fujiapple852, line 1: Would it be possible to store the snapshots _without_ stripping whitespace and only do so when comparing?  Not a deal breaker for using insta, but it would be nice.

crates/trippy/src/main.rs
```diff
@@ -485,20 +485,39 @@ mod tests {
...
-    #[test_case(include_str!("../test_resources/config/completions_fish.txt"), &shell_completions(Shell::Fish).unwrap(); "generate fish shell completions")]
-    #[test_case(include_str!("../test_resources/config/completions_powershell.txt"), &shell_completions(Shell::PowerShell).unwrap(); "generate powershell shell completions")]
-    #[test_case(include_str!("../test_resources/config/completions_zsh.txt"), &shell_completions(Shell::Zsh).unwrap(); "generate zsh shell completions")]
-    #[test_case(include_str!("../test_resources/config/trip.1"), &man_page().unwrap(); "generate man page")]
-    fn test_output(expected: &str, actual: &str) {
-        if remove_whitespace(actual.to_string()) != remove_whitespace(expected.to_string()) {
-            pretty_assertions::assert_eq!(actual, expected);
+    #[macro_export]
+    macro_rules! set_snapshot_settings {
+        ($info:expr, $($expr:expr),*) => {
```
fujiapple852, line 490: nit: `$expr:expr` is a bit confusing, I guess this could be called the `suffix`?

crates/trippy/src/main.rs
```diff
@@ -485,20 +485,39 @@ mod tests {
...
+
+            // Set snapshot suffix
+            settings.set_snapshot_suffix(format!($($expr,)*));
+
+            // Set snapshot storage location
+            let cwd = std::env::current_dir().expect("failed to get cwd");
+            let path = cwd.join("tests").join("resources").join("snapshots");
+            settings.set_snapshot_path(path);
+
+            let _guard = settings.bind_to_scope();
```
fujiapple852, line 504: Could we use `insta::with_settings!` for this?  It would mean passing the test body to the macro as an argument but I don't think that would be an issue?  Again, not a deal breaker.

crates/trippy/tests/resources/snapshots/trip__config__tests__compare_snapshot@trip --help.snap
fujiapple852: Also the fact it is named after the `compare_snapshot` function instead of the `test_help` in this case is a little unfortunate.

crates/trippy/Cargo.toml
c-git: Ok no problem will remove. I'm not sure why I didn't get the warning.

crates/trippy/tests/resources/snapshots/trip__config__tests__compare_snapshot@trip --help.snap
c-git: > Can we control the file names? I'm a little uneasy about filenames with spaces.

Yeah we control the filenames. I was just keeping it simple and using the command as the filename but it's trivial to change. Do you want me to use the same name without the spaces?

> Also the fact it is named after the compare_snapshot function instead of the test_help in this case is a little unfortunate.

That one is a bit more challenging because it uses the name where the assert happens. We "can" do it in each function but we end up with more boilerplate. When the test fails the test name will still show as `test_help`. If we wanted to it is something we can "fix" but not sure it's worth the effort. Not so much to change it but to maintain it.

crates/trippy/tests/resources/snapshots/trip__config__tests__compare_snapshot@trip -h.snap
c-git: I don't think so because the comparison is done against what is saved in the file. So we'd have to solve the width issue if we want to use it without the whitespace stripped in the file.

crates/trippy/src/main.rs
c-git: Yeah happy to change that. I didn't have a better name at the time so I just left the name from where I copied it.

crates/trippy/src/main.rs
c-git: Previously before I saw the example macro on their website I used to actually just have a [function](https://github.com/rust-practice/cargo-leet/blob/5f02af149192f151afaad7d288ec80e86c503b53/src/tool/core/helpers/local_store.rs#L59) that returned the settings. But I found the macro more ergonomic, but that's pretty subjective so I'm ok with going with what you prefer. Would you prefer not using a macro?

c-git: Well you said it would be nice so does that mean you're prefer not use the macro (Sorry just wanted to clarify what you want)?

.gitignore
```diff
@@ -1,4 +1,5 @@
 /target
 .idea
 .DS_Store
-.vscode/launch.json
\ No newline at end of file
+.vscode/launch.json
+*.snap.new
```
fujiapple852, line 5: @c-git I made this more specific, which I think is fine?

crates/trippy/src/config.rs
```diff
@@ -1695,11 +1714,6 @@ mod tests {
         )?)
     }
 
-    fn remove_whitespace(mut s: String) -> String {
-        s.retain(|c| !c.is_whitespace());
-        s
-    }
```
fujiapple852, line 1701: Now we have the `insta` fn as well I made the one in main::tests public to avoid repeating both of these

crates/trippy/src/main.rs
```diff
@@ -485,21 +485,29 @@ mod tests {
...
+        });
+    }
+
+    pub fn insta<F: FnOnce()>(name: &str, f: F) {
+        let mut settings = insta::Settings::new();
+        settings.set_snapshot_suffix(name.replace(' ', "_"));
+        settings.set_snapshot_path("../tests/resources/snapshots");
+        settings.set_omit_expression(true);
+        settings.bind(f);
     }
```
fujiapple852, line 508: @c-git I reworked this to avoid the need for a macro

crates/trippy/src/main.rs
```diff
@@ -485,21 +485,29 @@ mod tests {
...
+    #[test_case(&man_page().unwrap(), "generate man page"; "generate man page")]
+    fn test_output(actual: &str, name: &str) {
+        insta(name, || {
+            insta::assert_snapshot!(remove_whitespace(actual.to_string()));
+        });
+    }
+
+    pub fn insta<F: FnOnce()>(name: &str, f: F) {
+        let mut settings = insta::Settings::new();
+        settings.set_snapshot_suffix(name.replace(' ', "_"));
```
fujiapple852, line 504: @c-git escaping the name to avoid whitespace in filenames

crates/trippy/tests/resources/snapshots/trip__config__tests__compare_snapshot@trip --help.snap
fujiapple852: I fixed it to escape ' ' with "_" in the suffix

crates/trippy/src/main.rs
fujiapple852: Changed it to have an `insta` fn which takes a closure and removed the macro

crates/trippy/src/main.rs
```diff
@@ -485,21 +485,29 @@ mod tests {
...
+        settings.set_snapshot_path("../tests/resources/snapshots");
+        settings.set_omit_expression(true);
```
fujiapple852, line 506: @c-git added this as it seems to be just noise.

.gitignore
c-git: Yep that's better. In hindsight that's probably what it always should have been.

crates/trippy/src/config.rs
c-git: yeah makes sense

crates/trippy/src/main.rs
c-git: yeah that's fine.

crates/trippy/src/main.rs
c-git: Ok that's fair since all of them are called from the same place anyway.

crates/trippy/src/main.rs
```diff
@@ -485,21 +485,29 @@ mod tests {
...
+    fn test_output(actual: &str, name: &str) {
+        insta(name, || {
+            insta::assert_snapshot!(remove_whitespace(actual.to_string()));
+        });
+    }
+
+    pub fn insta<F: FnOnce()>(name: &str, f: F) {
+        let mut settings = insta::Settings::new();
+        settings.set_snapshot_suffix(name.replace(' ', "_"));
+        settings.set_snapshot_path("../tests/resources/snapshots");
```
c-git, line 505: I do have some reservations regarding the relative path use here but so far haven't seen it cause a problem. Did you dislike the absolute path approach?

crates/trippy/src/main.rs
c-git: Yep looks good.

c-git (commented):
Yeah looks good to me.

crates/trippy/src/main.rs
c-git: Since it's only called from one place I think it's fine as all of the users pass their use case as a closure.

PR #1145, opened 2024-05-16
Split Trippy into multiple crates

### LINKED ISSUE #1141: Separate lib and bin dependencies
Separate dependencies of trippy's public library components and it's CLI binary tool. This way, when trying to use trippy within another rust project, I can `cargo add trippy` without its version of `clap` (or any of the other TUI/CLI related dependencies) clashing with my own.

I'm not sure there is an alternative, besides me not using trippy as a library.

**Additional context**
I'm trying to use trippy's tracer within another rust project.

fujiapple852:
Closes #1141


Dockerfile
```diff
@@ -1,15 +1,32 @@
...
+
+# copy the actual application code and build
+COPY crates/trippy/src /app/crates/trippy/src
+COPY crates/trippy-core/src /app/crates/trippy-core/src
+COPY crates/trippy-dns/src /app/crates/trippy-dns/src
+COPY crates/trippy-privilege/src /app/crates/trippy-privilege/src
 COPY trippy-config-sample.toml /app
 COPY README.md /app
-COPY LICENSE /app
+RUN cargo clean --release --target=x86_64-unknown-linux-musl -p trippy-core -p trippy-dns -p trippy-privilege
```
fujiapple852, line 29: I don't know why this is needed, but it is.

README.md
```diff
@@ -8,7 +8,7 @@
 Trippy combines the functionality of traceroute and ping and is designed to assist with the analysis of networking
 issues.
 
-<img src="assets/0.8.0/trippy.gif" alt="trippy"/>
+<img src="https://raw.githubusercontent.com/fujiapple852/trippy/master/assets/0.8.0/trippy.gif" alt="trippy"/>
```
fujiapple852, line 11: This has to be an absolute url as the assets (gifs) are not in the `trippy` crate any more.  I'm hopeful that this will be rendered correctly by crates.io / docs.rs / lib.rs etc.

crates/trippy/src/lib.rs
```diff
@@ -0,0 +1,2 @@
+#![allow(rustdoc::broken_intra_doc_links, rustdoc::bare_urls)]
+#![doc = include_str!("../../../README.md")]
```
fujiapple852, line 2: I don't think this is needed, strictly speaking, as this is a binary crate.  Adding this will allow it to appear in docs.rs however.

.github/workflows/ci.yml
```diff
@@ -121,7 +121,7 @@ jobs:
         if: startsWith(matrix.build, 'windows')
         shell: bash
         run: |
-          cp "tests/resources/wintun.dll" "."
+          cp "crates/trippy-core/tests/resources/wintun.dll" "./crates/trippy-core/"
```
fujiapple852, line 124: Changed as the simulation test now runs from the `trippy-core` crate, so the `wintun.dll` must be copied there.

.github/workflows/ci.yml
```diff
@@ -194,6 +194,12 @@ jobs:
...
\ No newline at end of file
+        run: cargo install --git https://github.com/foresterre/cargo-msrv.git cargo-msrv
+      - name: check msrv for trippy
+        run: cargo msrv verify --output-format json --manifest-path crates/trippy/Cargo.toml -- cargo check
+      - name: check msrv for trippy-core
+        run: cargo msrv verify --output-format json --manifest-path crates/trippy-core/Cargo.toml -- cargo check
+      - name: check msrv for trippy-dns
+        run: cargo msrv verify --output-format json --manifest-path crates/trippy-dns/Cargo.toml -- cargo check
+      - name: check msrv for trippy-privilege
+        run: cargo msrv verify --output-format json --manifest-path crates/trippy-privilege/Cargo.toml -- cargo check
```
fujiapple852, line 205: Switched to master for cargo-msrv to get (partial) support for virtual workspaces, but requires it be run once per crate.

Dockerfile
```diff
@@ -1,15 +1,32 @@
-FROM rust:1.74 as build-env
+FROM rust:1.75 as build-env
```
fujiapple852, line 1: Bumped Dockerfile to MSRV, was a bug.

crates/trippy-core/Cargo.toml
```diff
@@ -0,0 +1,61 @@
...
+keywords = ["traceroute", "ping", "icmp"]
+categories = ["network-programming"]
+
+[dependencies]
+derive_more.workspace = true
+thiserror.workspace = true
+tracing.workspace = true
+itertools.workspace = true
+arrayvec = { version = "0.7.4", default-features = false }
+socket2 = { version = "0.5.7", features = [ "all" ] }
```
fujiapple852, line 22: Bug fix: added feature `all` which is required.  It used to work by accident (another dependancy added that feature) but failed if you build `trippy-core` only in the virtual workspace.

crates/trippy-core/src/net/platform/unix.rs
```diff
@@ -190,7 +190,7 @@ mod socket {
             })
         }
 
-        pub(super) fn new_raw_ipv4(protocol: Protocol) -> IoResult<Self> {
+        pub fn new_raw_ipv4(protocol: Protocol) -> IoResult<Self> {
```
fujiapple852, line 193: TODO: I didn't mean to change that

crates/trippy-dns/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+readme.workspace = true
+license.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+keywords = ["traceroute", "ping", "icmp"]
+categories = ["network-programming"]
+
+[dependencies]
+thiserror.workspace = true
+anyhow.workspace = true
```
fujiapple852, line 18: TODO: shouldn't be using anyhow in a library crate

crates/trippy-core/Cargo.toml
```diff
@@ -0,0 +1,61 @@
...
+hex-literal = "0.4.1"
+rand = "0.8.5"
+tokio = { version = "1.37.0", features = [ "full" ] }
+tokio-util = "0.7.11"
+ipnetwork = "0.20.0"
+mockall = "0.12.1"
+tracing-subscriber = "0.3.18"
+
+# see https://github.com/meh/rust-tun/pull/74
+[target.'cfg(any(target_os = "macos", target_os = "linux", target_os = "windows"))'.dev-dependencies]
```
c-git, line 56: Is there a reason not to just include this in the dev dependencies? It sees to be specified for all targets except WASM that I can think of.

c-git (approved):
Overall very happy with this PR. I did leave a few comments but none of them are show stoppers. I can try to do the removal of anyhow if you'd like. I've never done it before but I can look at how it's done in the privilege crate and try to pattern on that.

Dockerfile
wgreenberg: the dependency caching step creates empty dummy libraries with the same names as the trippy crates. then, when you copy the production code in, they try (and fail) to import modules from those empty dummy libraries -- hence why you need to delete the build artifacts.

fujiapple852: @wgreenberg I'd have expected that to work though?  The final build step (after copying in the real code) should rebuild the library crates as their src code has changed but it seems it doesn't "notice" that.

This approach certainly works for a single crate (without a virtual workspace) as that is what it does today: https://github.com/fujiapple852/trippy/blob/74c2c6a5a1d5f019858de78c83a7f77c8dbc0b4d/Dockerfile#L7-L13

So I can only assume a bug?

I did try disabling incremental compilation (`CARGO_INCREMENTAL=0`) in case it was the cause but that didn't seem to help.

README.md
fujiapple852: Yes not a blocker for merging, we can test separately and adjust if it doesn't work.

.github/workflows/ci.yml
fujiapple852: Good idea, added.

crates/README.md
```diff
@@ -0,0 +1,42 @@
...
+```
+
+### `trippy-core`
+
+A library crate providing the core Trippy tracing functionality. This crate is used by the Trippy application and is
+the crate you would use if you wish to provide the Trippy tracing functionality in your own application.
+
+```toml
+[dependencies]
+trippy-core = "0.10.0"
```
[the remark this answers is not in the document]
fujiapple852: Good idea, changed.

crates/trippy-core/Cargo.toml
fujiapple852: From memory it failed on some build targets in CI (netbsd and freebsd I think).

crates/trippy-privilege/src/lib.rs
```diff
@@ -1,8 +1,91 @@
...
+//! } else {
+//!     println!("You do not always need privileges to send ICMP packets.");
+//! }
+//! ```
+//!
+//! Drop all privileges:
+//!
+//! ```no_run
+//! # use trippy_privilege::Privilege;
+//! Privilege::drop_privileges().unwrap();
```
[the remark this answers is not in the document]
fujiapple852: Yes I think it may be safe to let this doctest run, changed.

crates/trippy-privilege/src/lib.rs
```diff
@@ -1,8 +1,91 @@
...
+//! ```
+#![warn(clippy::all, clippy::pedantic, clippy::nursery, rust_2018_idioms)]
+#![allow(clippy::missing_const_for_fn, clippy::missing_errors_doc)]
+#![deny(unsafe_code)]
```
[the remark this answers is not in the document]
fujiapple852: Personal preference mainly, I think in Rust the Result/Error types are already clear.  Adding an `# Errors` section to every fallible function (most...) is tedious to write and veery likely to fall out of sync with reality.  As the old saying goes, "code is the best documentation".

crates/trippy-privilege/src/lib.rs
```diff
@@ -1,8 +1,91 @@
...
+//! - On other Unix platforms this is a no-op
+//! - On Windows this is a no-op
+//!
+//! # Examples
+//!
+//! Acquire the required privileges if we can:
+//!
+//! ```no_run
+//! # use trippy_privilege::Privilege;
+//! Privilege::acquire_privileges().unwrap();
```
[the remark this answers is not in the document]
fujiapple852: Good call, changed.

crates/trippy-dns/Cargo.toml
fujiapple852: fixed

crates/trippy/Cargo.toml
```diff
@@ -0,0 +1,62 @@
...
+[dev-dependencies]
+test-case.workspace = true
+serde_yaml.workspace = true
+pretty_assertions = "1.4.0"
+
+# cargo-generate-rpm dependencies
+[package.metadata.generate-rpm]
+assets = [
+  { source = "target/release/trip", dest = "/usr/bin/trip", mode = "755" },
+]
```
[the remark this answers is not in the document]
fujiapple852: I'll have to test it by doing a release (from my testing shadow repo).  I suspect this can be replaced with a few command line arguments added to the GitHub action step.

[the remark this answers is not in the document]
fujiapple852: Fixed by adding `-p crates/trippy` to the `generate-rpm` command.  

Aside: `cargo-generate-rpm` [supports](https://github.com/cat-in-136/cargo-generate-rpm?tab=readme-ov-file#overwrite-configuration) inline toml configuration (via the `-s` flag) but I couldn't get it to work for the needed which has an array of nested objects:

```toml
assets = [
  { source = "target/release/trip", dest = "/usr/bin/trip", mode = "755" },
]
```

crates/trippy-privilege/src/lib.rs
```diff
@@ -1,8 +1,91 @@
...
+//! for sending ICMP packets via RAW and `IPPROTO_ICMP` sockets.
+//!
+//! `Privilege::acquire_privileges()`:
+//!
+//! - On Linux we check if `CAP_NET_RAW` is in the permitted set and if so raise
+//! it to the effective set
+//! - On other Unix platforms this is a no-op
+//! - On Windows this is a no-op
+//!
+//! `Privilege::has_privileges` (obtained via `Privilege::discover()`):
```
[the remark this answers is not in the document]
fujiapple852: Yes, fixed

Dockerfile
fujiapple852: Resolving this for now as it does not block merging, but it would be nice to understand what the issue is here and either fix it our report a bug.

PR #1209, opened 2024-07-01
Refactor tweaks

fujiapple852:


fujiapple852:
This is technically a breaking change (removing a parameter from `State::hops()`), however prior to version 0.11.0 Trippy has no official public API so there is no need to call this out in the CHANGELOG.

PR #1216, opened 2024-07-13
NAT detection for IPv4/UDP/Dublin

### LINKED ISSUE #1104: Detect NAT for `udp/dublin` strategy
The `udp/dublin` strategy makes it possible to detect NAT(s) that occur along the path of a given flow.  Some introductory information on the technique is given at https://dublin-traceroute.net/ and the details can be studied in the [code](https://github.com/insomniacslk/dublin-traceroute/blob/a92118d93fd1fa7bdb827e741dd848b7f7083a1e/go/dublintraceroute/probes/probev4/udpv4.go#L308).
 
For `udp/dublin` we maintain a stable udp checksum for each probe in a single round as the sequence number is encoded in the ip `identifier` field and so the virtual header is constant.  Therefore, if the udp checksum observed in the nested (original datagram) of an ICMP response differs then we know that NAT has occurred at some point along the path.

By comparing the udp checksum for all probes in a given round, specifically checking for changes in the checksum between hops which responded in the round, we can identify at which hop(s) NAT occurred and report this to the user.

Note: this does not work for `udp/paris` as the checksum differs for each probe in a round, as the checksum is used to encode the sequence number.  Indeed, `udp/paris` does not work at all when a path traverses a NAT.  Likewise, this does not work for `udp/classic` as the udp checksum is not stable between probes in a round as the sequence is recorded in the src or dest port, which are part of the virtual header used to compute the udp checksum.

In the following example (for `udp/dublin` over 1 round) all the outgoing udp packets have checksum `0xe5fc`.  The checksum observed (in the udp original datagram payload of the icmp response) is `0xe5fc` for the first hop (seq 33000) and there is no response for the probe at the 2nd hop (seq=33001).  

The 3rd hop (seq=33002) has a different observed checksum of `0x8eb3` which indicates that this hop performed a NAT.  Note that the NAT may have occurred at the previous hop (seq=33001) but as that did not respond in this round it is impossible to know.

The 4th hop (seq=33003) has the same checksum as the prior hop, `0x8eb3`, and so whilst this differs from the initial checksum (i.e. `0xe5fc`) this does not indicate an _additional_ NAT.  To say it another way, the change in checksum due to NAT "carries forward" throughout all remaining hops.

The 5th hop (seq=33004) shows a new checksum, `0xd993`, which differs from the previous hop and so again we can conclude that NAT occurs at this hop.  

```
Sent: seq=33000 checksum=0xe5fc
Recv: seq=33000 checksum=0xe5fc

Sent: seq=33001 checksum=0xe5fc
Recv: none

Sent: seq=33002 checksum=0xe5fc
Recv: seq=33002 checksum=0x8eb3 <- NAT #1

Sent: seq=33003 checksum=0xe5fc
Recv: seq=33003 checksum=0x8eb3

Sent: seq=33004 checksum=0xe5fc
Recv: seq=33004 checksum=0xd993 <- NAT #2
```

### LINKED ISSUE #1219: Add NAT status custom column
Add a custom column to indicate if NAT has been is detected for a given hop

fujiapple852:
Closes #1104 
Closes #1219


fujiapple852:
Cannot currently detect NAT at the very first hop.

fujiapple852:
Updated impl which now compares the expected UDP checksum vs the actual UDP checksum for the original datagram nested in the payload of the ICMP packet.  This should allow detecting NAT for the 1st hop (when there is no previous hop checksum to compare with).  Adds to the complexity a fair bit.

PR #1231, opened 2024-07-23
refactor(core): introduce `IPv4` and `IPv6` config structs

fujiapple852:


crates/trippy-core/src/net/ipv6.rs
```diff
@@ -47,252 +47,364 @@ const MIN_PACKET_SIZE_UDP: usize =
...
-    process_result(local_addr, socket.bind(local_addr))?;
-    socket.set_unicast_hops_v6(probe.ttl.0)?;
-    socket.send_to(payload, remote_addr)?;
-    Ok(())
-}
+    #[instrument(skip(self, probe))]
+    fn dispatch_udp_probe_non_raw<S: Socket>(
+        &self,
+        probe: Probe,
+        // src_addr: Ipv6Addr,
```
fujiapple852, line 164: Commented out code to remove

PR #1326, opened 2024-10-01
fix: rename IP to Ip

### LINKED ISSUE #1327: `tui-address-mode` does not accept `ip`
**Describe the bug**

Instead of `tui-address-mode` accepting `ip` as a valid value it expects `i-p`.

```
trip 8.8.8.8
Error: TOML parse error at line 243, column 20
    |
243 | tui-address-mode = "ip"
    |                    ^^^^
unknown variant `ip`, expected one of `i-p`, `host`, `both`
```

**To Reproduce**

 - Create a config file using `trip --print-config-template > trippy.toml`
 - Then edit `tui-address-mode` to be `ip`.
 - Then run trippy `trip 8.8.8.8`

**Expected behavior**

Expected mode to be set to `ip` with trip running

**Environment Info**
- OS:  Ubuntu
- Trippy version:  trip 0.11.0
- Installation method:  cargo
- Terminal / Console:  GNOME Terminal - Version 3.44.0 for GNOME 42

c-git:
Needed because kebab-case for `IP` is `i-p` but example toml file shows `ip`
so renamed to `Ip` so that kebab version is `ip`

Closes #1327

### DIFF
--- a/crates/trippy-tui/src/config.rs
+++ b/crates/trippy-tui/src/config.rs
@@ -127,7 +127,7 @@ impl From<MultipathStrategy> for MultipathStrategyConfig {
 #[serde(rename_all = "kebab-case")]
 pub enum AddressMode {
     /// Show IP address only.
-    IP,
+    Ip,
     /// Show reverse-lookup DNS hostname only.
     Host,
     /// Show both IP address and reverse-lookup DNS hostname.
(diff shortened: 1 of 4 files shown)


c-git:
I ran into an issue when trying to set my default to always be `ip` (as shown below), and figured it was faster to do a PR than to explain to ask if that's the way you wanted to resolve the problem.

```
trip 8.8.8.8
Error: TOML parse error at line 243, column 20
    |
243 | tui-address-mode = "ip"
    |                    ^^^^
unknown variant `ip`, expected one of `i-p`, `host`, `both`
```

fujiapple852:
Good catch!

The fix LGTM.  You could also have added `[serde(rename = “ip”)]` on the `IP` variant but I think I prefer your solution. 

I think this is worth a mention in the next release note so can you create a bug issue and link to it here as “Closes #xxxx”.

c-git:
Ok good idea, will do. It's also in keeping with the API naming [guidelines](https://rust-lang.github.io/api-guidelines/naming.html). 

> In UpperCamelCase, acronyms and contractions of compound words count as one word: use Uuid rather than UUID, Usize rather than USize or Stdin rather than StdIn. In snake_case, acronyms and contractions are lower-cased: is_xid_start.

PR #1334, opened 2024-10-03
feat(tui): add Portuguese (`pt`) locale

fujiapple852:


crates/trippy-tui/locales/app.yml
```diff
@@ -3,439 +3,549 @@ trippy:
...
 target:
   en: "Target"
   fr: "Cible"
   tr: "Hedef"
+  pt: "Alvo"
 config:
   en: "Config"
   fr: "Config"
   tr: "Konfigurasyon"
+  pt: "Config"
```
ulissesf, line 61: Config can be used as a short version, but actual word is "Configuração"

crates/trippy-tui/locales/app.yml
```diff
@@ -3,439 +3,549 @@ trippy:
...
 protocol:
   en: "protocol"
   fr: "protocole"
   tr: "protokol"
+  pt: "protocolo"
 as-info:
   en: "as-info"
   fr: "info-as"
   tr: "AS bilgisi"
+  pt: "info-as"
```
ulissesf, line 76: I think "as" is an acronym here? So some type of information? Can this field be longer than the "en" text or not?

crates/trippy-tui/locales/app.yml
```diff
@@ -3,439 +3,549 @@ trippy:
...
 details:
   en: "details"
   fr: "détails"
   tr: "ayrıntılar"
+  pt: "detalhes"
 max-hosts:
   en: "max-hosts"
   fr: "nombre d'hôtes maximum"
   tr: "maksimum hostlar"
+  pt: "max-hosts"
```
ulissesf, line 86: You can shorten it to "max-hosts" in Portuguese in order to say "maximum number of hosts". The Portuguese translation for the longer text would be "máximo número de hosts".

crates/trippy-tui/locales/app.yml
```diff
@@ -3,439 +3,549 @@ trippy:
...
 dns_failed:
   en: "Failed"
   fr: "Échec"
   tr: "Başarısız"
+  pt: "Falhou"
 dns_timeout:
   en: "Timeout"
   fr: "Délai dépassé"
   tr: "Zaman aşımı"
+  pt: "Tempo esgotado"
```
ulissesf, line 266: Is "Timeout" being used when something actually times out or to display a timeout (time limit)? If the former, then "Tempo esgotado" is correct, if the latter then "Limite de tempo" is better.

crates/trippy-tui/locales/app.yml
fujiapple852: Ok, will keep as "Config" then.

crates/trippy-tui/locales/app.yml
fujiapple852: I'm trying to get rid of this one entirely as it is very long and many languages and also very unclear to users what it even means!

crates/trippy-tui/locales/app.yml
fujiapple852: It is the former; this will be shown if a time limit expires when trying to lookup the hostname associated with a given IP address.  So will keep it as-is.

crates/trippy-tui/locales/app.yml
fujiapple852: I'm also removing this field so you can ignore it, so will leave as it is for now.

fujiapple852:
Here is what it looks like @ulissesf:

Screenshot 2024-10-18 at 10 50 44 PM

I think we're good to merge, thanks again for your time!

fujiapple852:
cc @0323pin for `pt` translation.

PR #1339, opened 2024-10-04
doc: make as searchable

c-git:
Makes it easier for users to lookup as AS is hard to search for


fujiapple852:
My only comment is on the `cmd.rs` change, as this makes `--tui-as-mode` the new longest line output by `trip -h`:

```shell
      --tui-as-mode <TUI_AS_MODE>                          How to render Anonymous System (AS) information [default: asn] [possible values: asn, prefix, country-code, registry, allocated, name]
```

The output is a bit of inconsistent at the moment, we don't define some of the terms here (i.e. MPLS) though we do define some (i.e. ECMP).  On balance I _think_ I favour brevity for command line help output, but I'm persuadable either way.

Fine with everything else.

c-git:
I don't feel strongly about changing that one. I don't really mind not defining MPLS but I do mind for AS. The reason is MPLS is very easy to Google while AS is a lot harder because "as" is a word.

fujiapple852:
Tweaked slightly to standardize on `anonymous system (AS)` (just a change to the casing).  I also rebased and squashed the last two commits.

fujiapple852:
Seems I had some typos ("Anonymous System (AS)") in there as well, all fixed.

PR #1353, opened 2024-10-15
Add `expand-privacy` and `contract-privacy` bindings

### LINKED ISSUE #1347: [BREAKING CHANGE] Replace `toggle-privacy` key binding with `expand-privacy` and  `contract-privacy`
Add `expand-privacy` and `contract-privacy` key bindings which increase and decrease `tui-privacy-max-ttl`.  The value is clamped in the range 0 (i.e. off) and the current maximum ttl of the trace.

## Deprecation

This is a breaking change as the existing `toggle-privacy` key binding has been deprecated.

If used from a configuration file it will result in an error:

```shell
$ sudo ./target/debug/trip 1.1.1.1
Error: toggle-privacy in [bindings] section is deprecated, use expand-privacy and contract-privacy instead
```

If used as a command line parameter (unlikely) it will result in an error:

```shell
$ sudo ./target/debug/trip 1.1.1.1 --tui-key-bindings toggle-privacy=o
error: invalid value 'toggle-privacy=o' for '--tui-key-bindings <TUI_KEY_BINDINGS>': toggle-privacy is deprecated, use expand-privacy and contract-privacy instead

For more information, try '--help'.
```

fujiapple852:
Closes #1347


crates/trippy-tui/src/config.rs
```diff
@@ -811,11 +811,16 @@ const fn cfg_layer_bool_flag(fst: bool, snd: Option<bool>, default: bool) -> boo
...
-fn validate_deprecated(cfg_file_tui: &ConfigTui) -> anyhow::Result<()> {
+fn validate_deprecated(
+    cfg_file_tui: &ConfigTui,
+    cfg_file_tui_bindings: &ConfigBindings,
+) -> anyhow::Result<()> {
     if cfg_file_tui.deprecated_tui_max_samples.is_some() {
         Err(anyhow!("tui-max-samples in [tui] section is deprecated, use max-samples in [strategy] section instead"))
     } else if cfg_file_tui.deprecated_tui_max_flows.is_some() {
         Err(anyhow!("tui-max-flows in [tui] section is deprecated, use max-flows in [strategy] section instead"))
+    } else if cfg_file_tui_bindings.deprecated_toggle_privacy.is_some() {
```
c-git, line 822: I'm wondering if this would be easier to read as a match statement?

c-git (approved):
Looks good overall. Will install and test.

crates/trippy-tui/src/config.rs
fujiapple852: I don't think so:

```rust
    match (
        cfg_file_tui.deprecated_tui_max_samples.is_some(),
        cfg_file_tui.deprecated_tui_max_flows.is_some(),
        cfg_file_tui_bindings.deprecated_toggle_privacy.is_some(),
    ) {
        (true, _, _) => Err(anyhow!("tui-max-samples in [tui] section is deprecated, use max-samples in [strategy] section instead")),
        (_, true, _) => Err(anyhow!("tui-max-flows in [tui] section is deprecated, use max-flows in [strategy] section instead")),
        (_, _, true) => Err(anyhow!("toggle-privacy in [bindings] section is deprecated, use expand-privacy and contract-privacy instead")),
        _ => Ok(()),
    }
```

PR #1374, opened 2024-11-05
fix(tui): cursor is not moved to the bottom on exit when using `--tui-preserve-screen`

### LINKED ISSUE #1375: cursor is not moved to the bottom on exit when using `--tui-preserve-screen`
cursor is not moved to the bottom on exit when using `--tui-preserve-screen`

### LINKED ISSUE #1382: add `quit-preserve-screen` key binding
Trippy already supports `--tui-preserve-screen` to preserve the screen when the application exits but there is no way to enable this once the application has started.

Add a new Tui command and key binding (default to `shift+q`) to quit the application and preserve the screen.

fujiapple852:
Closes #1375 
Closes #1382


c-git (approved):
Works well for me in both bash and zsh

PR #1421, opened 2024-12-10
update `hickory` to `0.25.0-alpha.4`

fujiapple852:


fujiapple852:
The `hickory-resolver` library will be divested: https://github.com/fujiapple852/trippy/issues/1706

PR #1424, opened 2024-12-11
Replace usage of YAML with TOML

### LINKED ISSUE #1416: Get rid of (serde_)y(a)ml
serde_yaml is unmaintained, so trippy switched to the serde_yml fork, but it has a worrying code quality ([hardcoding `c_char` as `i8`](https://github.com/sebastienrousseau/libyml/pull/6)). YAML is a confusing format to begin with. So, it'd be nice if we could switch to TOML, which is already a dependency. Currently (almost) all usage of YAML is in test data. For example, 

https://github.com/fujiapple852/trippy/blob/e2c7726dc924d49c7a98461365f985e88076f394/crates/trippy-core/tests/resources/state/nat.yaml#L1-L22

This could be translated to:

```toml
largest_ttl = 3

[[rounds]]
probes = [
  "1 C 333 10.1.0.1 0 12340 80 43012 43012",
  "2 C 777 10.1.0.2 1 12340 80 20544 20544",
  "3 C 778 10.1.0.3 2 12340 80 20544 20544"
]

[[rounds]]
probes = [
  "1 C 123 10.1.0.1 3 12340 80 43012 43012",
  "2 C 788 10.1.0.2 4 12340 80 20544 20544",
  "3 C 789 10.1.0.3 5 12340 80 20544 20544"
]

[[rounds]]
probes = [
  "1 C 123 10.1.0.1 6 12340 80 43012 43012",
  "2 C 780 10.1.0.2 7 12340 80 20544 20544",
  "3 C 781 10.1.0.3 8 12340 80 20544 20544"
]

[[expected.hops]]
ttl = 1
last_nat_status = "no_nat"

[[expected.hops]]
ttl = 2
last_nat_status = "nat"

[[expected.hops]]
ttl = 3
last_nat_status = "no_nat"
```

Some files use the tag feature of YAML e.g. `!SingleHost`, which presumably could be replaced with an optional key in TOML and some processing in the code.

The usage other than "almost all" is in i18n strings, introduced by rust-i18n, which is not packaged in Debian and "polyfilled" with an [ad-hoc patch](https://salsa.debian.org/debian/trippy/-/blob/debian/sid/debian/patches/ad-hoc-i18n.patch) to deliver the new functionality ASAP without seeing rust-i18n packages through NEW queue. Arguably, it could be replaced with [i18n-embed](https://lib.rs/crates/i18n-embed), which supports the Fluent translation system, as well as the old gettext system, both very popular and widely supported. It has built-in locale resolution through the [locale_config](https://lib.rs/crates/locale_config) crate. But this is probably too big a change, and a too abrupt one shortly after i18n is released.

Note this is strongly influenced by my PoV as a downstream package maintainer, focusing on maintainability (serde_yml) and avoiding unnecessary dependencies (serde_yml, rust-i18n), which are not always the biggest concerns for upstream developers.

I'm willing to implement the change if either is considered acceptable.

nc7s:
Resolves #1416. It also discussed i18n, but that is probably better suited as a separate PR.


crates/trippy-core/Cargo.toml
```diff
@@ -49,10 +49,10 @@ ipnetwork.workspace = true
 mockall.workspace = true
 rand.workspace = true
 serde = { workspace = true, default-features = false, features = [ "derive" ] }
-serde_yml.workspace = true
```
fujiapple852, line 52: @nc7s you can remove this from the workspace dependencies as well as it is now unused

Cargo.lock
```diff
@@ -1825,7 +1815,7 @@ dependencies = [
  "rust-i18n-support",
  "serde",
  "serde_json",
- "serde_yml 0.0.11",
+ "serde_yml",
```
fujiapple852, line 1818: @nc7s it looks like `rust-i18n` uses `serde_yml` unconditionally (not feature gated) so even after switching to use `toml` we still have the crate in the graph.

```
$ cargo tree -i serde_yml
serde_yml v0.0.11
├── rust-i18n-macro v3.1.2 (proc-macro)
│   └── rust-i18n v3.1.2
│       └── trippy-tui v0.13.0-dev (trippy/crates/trippy-tui)
│           └── trippy v0.13.0-dev (trippy/crates/trippy)
│               ├── hello-world v0.1.0 (trippy/examples/hello-world)
│               └── traceroute v0.1.0 (trippy/examples/traceroute)
└── rust-i18n-support v3.1.2
    ├── rust-i18n v3.1.2 (*)
    └── rust-i18n-macro v3.1.2 (proc-macro) (*)
```

nc7s: Indeed, it would be removed entirely in the next PR to replace it with i18n-embed (or other, this is the one I'm experimenting).

fujiapple852:
LGTM

PR #1426, opened 2024-12-11
Replace rust-i18n with i18n-embed

nc7s:
As discussed in #1416. From there:

> I'm also tempted to just do something akin to your patch, as doing this translation is simple enough that pulling in dependencies may be overkill.

Probably, if strings are to be fairly simple and keep that way. Optimize with phf, even.

Anyway, the new code is here for comparison.

This is a very crude draft with quite some points up for debate: code style and placement, workaround for locale override with config value (there's a comment about this), structure of i18n resources, and MSRV (enabling `std::sync::LazyLock`, but actually not very relevant to i18n).

Good news is this does not significantly affect compiled size (even down a few kilos actually).


fujiapple852:
Thanks, I'll review this properly later today.

Once thing worth doing is re-introducing the `locale::t!` "wrapper" macro such that no changes are needed at the call site (i.e. everything in `frontend::render`).  

I tested this locally and it seemed to work ok:

```
#[macro_export]
macro_rules! t {
    ($($all:tt)*) => {
        i18n_embed_fl::fl!($crate::locale::LANGUAGE_LOADER, $($all)*)
    }
}
```

nc7s:
Indeed, that'll remove quite some noise. Though `fl!` returns `String`, so some invocations still need to be adjusted.

crates/trippy-tui/src/frontend/columns.rs
```diff
@@ -271,8 +271,11 @@ impl ColumnType {
     /// The name of the column in the current locale.
     #[allow(clippy::cognitive_complexity)]
     pub(self) fn name(&self) -> Cow<'_, str> {
+        if self == &Self::Ttl {
+            return "#".into();
+        }
         match self {
-            Self::Ttl => Cow::Borrowed("#"),
+            Self::Ttl => unreachable!(),
```
nc7s, line 278: Avoids introducing `.into()` to all `t!()` invocations at the cost of extra statements other than the match. I'll switch if you'd prefer.

crates/trippy-tui/src/frontend/columns.rs
```diff
@@ -426,6 +430,7 @@ mod tests {
     #[test_case(ColumnType::StdDev, "StDev")]
     #[test_case(ColumnType::Status, "Sts")]
     fn test_column_display_formatting(c: ColumnType, heading: &'static str) {
+        let _ = crate::locale::init(Some("en"));
```
nc7s, line 433: This (and the one below) is due to `i18n_embed::select()` being called in `crate::locale::init()`, which is to take `TrippyConfig::tui_locale` into account. Being called on each test invocation shouldn't matter much, it's called only once in the main program. Still, the aesthetics might not be favorable.

crates/trippy-tui/src/frontend/render/bar.rs
```diff
@@ -15,18 +15,18 @@ pub fn render(f: &mut Frame<'_>, rect: Rect, app: &TuiApp) {
...
-            t!("UDP"),
+            "UDP",
             app.tracer_config().data.multipath_strategy(),
         ),
         Protocol::Tcp => format!(
             "{}/{}",
             fmt_target_family(app.tracer_config().data.target_addr()),
-            t!("TCP"),
+            "TCP",
         ),
```
nc7s, line 30: Arguably these abbreviations are fixated and accepted in other languages, no need to translate?

crates/trippy-tui/src/frontend/columns.rs
fujiapple852: This got me thinking:

The `ColumnType::name` method here returns `Cow<'_, str>` only because that is what the `i18n-rust` crate macro returned.  Easy to fix, we just make the return type here (and a few other places) a String.  

However, does that mean that `i18n-embed` allocates an owned string every time? After digging through the code of the `i18n_embed_fl::fl!` macro it seems it ultimate calls this:

https://github.com/kellpossible/cargo-i18n/blob/master/i18n-embed/src/fluent.rs#L133-L136

```rust
/// Get a localized message referenced by the `message_id`.
pub fn get(&self, message_id: &str) -> String {
    self.get_args_fluent(message_id, None)
}
```

So the answer appears to be that it does allocate a new String for every translation, every time.

crates/trippy-tui/src/frontend/render/bar.rs
fujiapple852: I think this is fine, I put the following in the 0.12.0 release note:
> Many common abbreviated technical terms, such as IPv4 and ASN, are not translated.

I'd say if we are going to not attempt to translate these technical abbreviations then we should do so for all of them (so, everything in the current locale file where all languages use the same english abbreviation.  I can compile a list.  It feels like an orthogonal topic so I'd be tempted to do this as a separate PR, either before or after this one.

fujiapple852:
@nc7s I've pushed some WIP commits to this PR (we can refine these and squash later).

- wip: remove unused sys-locale crate 

No longer used so I removed it.  I haven't reviewed the replacement logic yet.

- wip: remove Cow from several methods

See my comment above re: removing `Cow` and what that reveals about the way the `i18n_embed` crate works.

- wip: use OnceLock instead of LazyLock

We can use `OnceLock` (stable on 1.70, Jun 2023) instead of `LazyLock` (1.80, Jul 2024) to avoid the MSRV bump, as the latter is _a little bit_ too recent (thinking about package maintainers here!).

- wip: move i18n into crates/trippy-tui and delete old locales

As the translations are strictly a TUI concern I'd prefer this to live in the `trippy-tui` crate.  It was quite fiddly to move, but it seems to work.  It seems `rust-embed` handles differently between debug and release builds but has a `"debug-embed"` to align them, which I briefly tried but didn't use.

I'm find it confusing to have a `i18n.toml` config file which defines the `assets_dir` and having to specify it in `#[folder = "i18n"]`.  I did look briefly if there was a way to avoid this but didn't see anything.

I also removed the old locale directory in this commit.

fujiapple852:
@nc7s on the fluent `ftl` files, do you know if it is possible to combine them into a single multi-language file?  This has pros and cons but is handy for generating placeholder translations for new terms for all languages quickly as it plays well with AI tools.

If not, can we name all file as `${lang}_ftl` and place them in a single directory? (aside: is there any significance to the name `trippy-tui.ftl` or is it arbitrary?)

Cargo.lock
fujiapple852, line 1: @nc7s I trust all of the new crates this brings in are already available/packaged for Debian?

crates/trippy-tui/i18n.toml
fujiapple852, line 1: As noted elsewhere, it would be nice if we didn't need this config file, especially as the `assets_dir` seems to also need to be specified elsewhere.  Not a showstopper if not.

crates/trippy-tui/src/frontend/columns.rs
```diff
@@ -270,9 +269,9 @@ impl Display for ColumnType {
 impl ColumnType {
     /// The name of the column in the current locale.
     #[allow(clippy::cognitive_complexity)]
-    pub(self) fn name(&self) -> Cow<'_, str> {
+    pub(self) fn name(self) -> String {
```
fujiapple852, line 272: @nc7s note the change in signature here, since the macro now already returns a `String`.

crates/trippy-tui/src/frontend/render/bar.rs
```diff
@@ -111,7 +110,7 @@ pub fn render(f: &mut Frame<'_>, rect: Rect, app: &TuiApp) {
     f.render_widget(left, rect);
 }
 
-fn fmt_privilege_mode(privilege_mode: PrivilegeMode) -> Cow<'static, str> {
+fn fmt_privilege_mode(privilege_mode: PrivilegeMode) -> String {
```
fujiapple852, line 113: @nc7s another place where I changed `Cow` to `String`.

crates/trippy-tui/src/frontend/render/splash.rs
```diff
@@ -25,13 +24,13 @@ pub fn render(f: &mut Frame<'_>, app: &TuiApp, rect: Rect) {
                 .fg(app.tui_config.theme.text),
         );
     #[allow(clippy::needless_raw_string_hashes)]
-    let splash: Vec<Cow<'static, str>> = vec![
+    let splash: Vec<String> = vec![
```
fujiapple852, line 27: @nc7s and another `Cow`->`String`

crates/trippy-tui/src/locale.rs
```diff
@@ -1,80 +1,38 @@
...
-        .map_or_else(|| FALLBACK_LOCALE, |lang| lang)
-        .to_string()
-}
-
-// A macro for translating a text string.
 #[macro_export]
 macro_rules! t {
     ($($all:tt)*) => {
-        rust_i18n::t!($($all)*)
+        i18n_embed_fl::fl!($crate::locale::LANGUAGE_LOADER.get_or_init(|| $crate::locale::fluent_language_loader!()), $($all)*)
```
fujiapple852, line 16: @nc7s I modified this to work with `OnceLock`.

crates/trippy-tui/src/print.rs
```diff
@@ -32,7 +34,11 @@ pub fn print_man_page() -> anyhow::Result<()> {
 }
 
 pub fn print_locales() {
-    println!("TUI locales: {}", available_locales().iter().join(", "));
+    let locales = LANGUAGE_LOADER
+        .get_or_init(|| fluent_language_loader!())
```
fujiapple852, line 38: @nc7s modified to work with `OnceLock`.

fujiapple852: aside: we may want to push this impl into the locale `module` to avoid exposing `LANGUAGE_LOADER` (we can make it `doc(hidden)` perhaps since it must be public).

crates/trippy-tui/src/locale.rs
```diff
@@ -1,80 +1,38 @@
...
-    } else {
-        set_locale_inner(FALLBACK_LOCALE);
-    }
-}
-
-/// Get the current locale.
-pub fn locale() -> String {
-    rust_i18n::locale().to_string()
-}
+pub use i18n_embed::fluent::fluent_language_loader;
```
fujiapple852, line 5: pub re-export as it is referenced by the macro

crates/trippy-tui/src/frontend/columns.rs
nc7s: `i18n_embed::fluent::FluentLanguageLoader::get_args_fluent()` ultimately calls [`fluient::bundle::FluentBundle::format_pattern()`](https://docs.rs/fluent/latest/fluent/bundle/struct.FluentBundle.html#method.format_pattern), which actually returns a `Cow<_, str>`. It then turns it `.into()` a `String`. This is a bit frustrating.

crates/trippy-tui/src/frontend/columns.rs
nc7s: https://github.com/fujiapple852/trippy/blob/98f3f8e24e4c1e83050dba1a45201d5ade85bf43/crates/trippy-tui/src/lib.rs#L23-L24

Nevertheless, it has to use a global lazy value, we can too, but I decided to go with `locale::init()` to reduce that a bit. (early optimization? heh)

Cargo.lock
nc7s: Yes.

crates/trippy-tui/i18n.toml
nc7s:
```
error: fl!() had a problem reading i18n config file "/home/i/Works/trippy/crates/trippy-tui/i18n.toml": Cannot read file "/home/i/Works/trippy/crates/trippy-tui/i18n.toml" in the current working directory Ok("/home/i/Works/trippy") because No such file or directory (os error 2).

         = help: Try creating the `i18n.toml` configuration file.

   --> crates/trippy-tui/src/locale.rs:16:9
    |
16  | ...   i18n_embed_fl::fl!($crate::locale::LANGUAGE_LOADER.get_or_init(|| $crate::locale::fluent_language_loader!()), $($all)*)
    |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
   ::: crates/trippy-tui/src/frontend/columns.rs:138:41
    |
138 |             Self::Shown => f.write_str(&t!("on")),
    |                                         -------- in this macro invocation
    |
    = note: this error originates in the macro `i18n_embed_fl::fl` which comes from the expansion of the macro `t` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Not so nice, unfortunately.

crates/trippy-tui/src/locale.rs
nc7s: This seems a bit all over the place since it's a macro. Maybe pack it into a function? (But then it's either a function call, or if a `const` function, still all over the place..)

nc7s:
Thanks for the detailed review and code.

> - wip: move i18n into crates/trippy-tui and delete old locales
> As the translations are strictly a TUI concern I'd prefer this to live in the trippy-tui crate. It was quite fiddly to move, but it seems to work. It seems rust-embed handles differently between debug and release builds but has a "debug-embed" to align them, which I briefly tried but didn't use.

I regret to admit it's because I simply followed its docs, which told me to put it in the root. :>

> I'm find it confusing to have a i18n.toml config file which defines the assets_dir and having to specify it in #[folder = "i18n"]. I did look briefly if there was a way to avoid this but didn't see anything.

The former is for the i18n-embed crate, the latter for rust-embed. The former does not actually _embed_ files itself, even with that word in the name.

> @nc7s on the fluent ftl files, do you know if it is possible to combine them into a single multi-language file? This has pros and cons but is handy for generating placeholder translations for new terms for all languages quickly as it plays well with AI tools.

As shown in rust-i18n docs:

```
# Use the fluent localization system.
[fluent]
# (Required) The path to the assets directory.
# The paths inside the assets directory should be structured like so:
# `assets_dir/{language}/{domain}.ftl`
assets_dir = "i18n"
```

I'm afraid we have to manually reimplement some of the set code to support an alternative layout, e.g. building `FluentBundle`s ourselves.

> If not, can we name all file as ${lang}_ftl and place them in a single directory? (aside: is there any significance to the name trippy-tui.ftl or is it arbitrary?)

As shown above, `trippy-tui.ftl` is just `{domain}.ftl`, which we do have control by manually building a `FluentLanguageLoader` (not using the `fluent_language_loader!` macro).

c-git:
Maybe this is a good use case to implement a few bench marks?

nc7s:
I used its thread safe version, `std::sync::LazyLock`, but it requires bumping MSRV to 1.80, @fujiapple852 worried it's too new and changed "back" to `std::sync::OnceLock`, which was introduced in 1.70, lower than current MSRV 1.76. It has the cost of lock, but it seems we need it for `Sync`:

```
error[E0277]: `OnceCell<FluentLanguageLoader>` cannot be shared between threads safely
  --> crates/trippy-tui/src/locale.rs:12:29
   |
12 | pub static LANGUAGE_LOADER: OnceCell<FluentLanguageLoader> = OnceCell::new();
   |                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `OnceCell<FluentLanguageLoader>` cannot be shared between threads safely
   |
   = help: the trait `Sync` is not implemented for `OnceCell<FluentLanguageLoader>`
   = note: if you want to do aliasing and mutation between multiple threads, use `std::sync::OnceLock` instead
   = note: shared static variables must have a type that implements `Sync`

For more information about this error, try `rustc --explain E0277`.
error: could not compile `trippy-tui` (lib) due to 1 previous error
```

c-git:
Oh sorry I guess I should have remember it needed to go across threads. Yes I think the lock cost is less than creating the string each time but we'd probably need to benchmark it to really know. And yes using OnceLock would also be fine.

nc7s:
Arguably we could `.leak()` a `String` into it, since the locale is almost always set only once per run, but we still need to figure out how to write into a `static` without `OnceLock` or the like.

c-git:
I think I missed something, why can't we use OnceLock?

nc7s:
Uh that was intended as an idea for further optimization, `OnceLock` is fine per se.

fujiapple852:
@nc7s I pushed another wip commit, this is mostly about making the i18n impl hidden (to the extent possible) inside the locale module.  I also added some tests.

crates/trippy-tui/src/frontend/columns.rs
```diff
@@ -426,6 +425,7 @@ mod tests {
     #[test_case(ColumnType::StdDev, "StDev")]
     #[test_case(ColumnType::Status, "Sts")]
     fn test_column_display_formatting(c: ColumnType, heading: &'static str) {
+        let _ = crate::locale::init(None);
```
nc7s, line 428: This is not gonna cut it, as system language is taken into account, so in an environment with e.g. `LANG=zh` it will break. The exact reason I wrote `init(Some("en"))`.

crates/trippy-tui/src/locale.rs
```diff
@@ -1,40 +1,82 @@
...
+    fn test_available_languages() {
+        assert_eq!(
+            available_languages().unwrap(),
+            vec!["de", "en", "es", "fr", "it", "pt", "ru", "sv", "tr", "zh"]
+        );
+    }
+
+    #[test]
+    fn test_init() {
+        assert_eq!(init(None).unwrap(), "en");
```
fujiapple852, line 72: Same here I guess

crates/trippy-tui/src/frontend/columns.rs
fujiapple852: I ran a similar test of trippy 0.12.0 vs this branch (build in release mode) and there is no significant difference in CPU usage on my system (`x86_64` MacBook Pro).  I also compared vs 0.11.0. The numbers are all within the margin of error.

0.11.0: ~1.5%
0.12.0: ~1.6%
0.13.0-dev: ~1.7%

So I don't think this is a real-world concern.

fujiapple852:
Closed in favour of https://github.com/fujiapple852/trippy/pull/1428

PR #1428, opened 2024-12-13
Replace rust-i18n with phf string maps and simple replaces

nc7s:
Alternative to #1426 for comparison. Strings are now `&'static str` as `phf::Map` entries, placeholder substitution is simply `str::replace`, no complex logic. Executable size is down 1MB (!).

Downsides are 1. build.rs, 2. `str::replace` is not enough for complex translations.

Currently both the `$kt:ident = $kt:expr` and `$kt:literal => $kv:expr` forms of `t!` are still supported, but I doubt if they really differ.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,7 +61,6 @@ petgraph = "0.6.5"
 pretty_assertions = "1.4.1"
 rand = "0.8.5"
 ratatui = "0.29.0"
-rust-i18n = "3.1.2"
 serde = { version = "1.0.201", default-features = false }
 serde_json = { version = "1.0.117", default-features = false }
 serde_with = { version = "3.11.0", default-features = false, features = ["macros"] }
--- a/crates/trippy-tui/src/locale.rs
+++ b/crates/trippy-tui/src/locale.rs
@@ -1,45 +1,131 @@
+use itertools::Itertools;
+use serde::Deserialize;
+use std::cell::RefCell;
+use std::collections::HashMap;
+use std::sync::OnceLock;
+
 const FALLBACK_LOCALE: &str = "en";
 
 /// Set the locale for the application.
 ///
-/// If the given locale is `None` or unsupported, the system locale is tried. If the system locale
-/// is not supported, the fallback locale is used.
+/// If the given locale is `None` the system locale is tried. If the system locale cannot be
+/// determined then the fallback locale is used.
 ///
-/// In both cases, the language part of the locale is used if the full locale is not supported.
-pub fn set_locale(locale: Option<&str>) {
-    if let Some(locale) = locale {
-        set_locale_inner(locale);
-    } else if let Some(locale) = sys_locale::get_locale().as_ref() {
-        set_locale_inner(locale);
+/// In all cases, the language part of the locale is used if the full locale is not supported.
+pub fn set_locale(locale: Option<&str>) -> String {
+    let new_locale = calculate_locale(locale, sys_locale::get_locale().as_deref());
+    store_locale(&new_locale);
+    new_locale
+}
+
+/// Get the available locales.
+pub fn available_locales() -> Vec<&'static str> {
+    data()
+        .0
+        .iter()
+        .flat_map(|(_, v)| v.0.keys().map(AsRef::as_ref))
+        .unique()
+        .sorted_unstable()
+        .collect::<Vec<_>>()
+}
+
+/// A macro to translate an item to the current locale.
+#[macro_export]
+macro_rules! t {
+    ($key:expr) => {
+        std::borrow::Cow::Borrowed($crate::locale::__translate($key))
+    };
+    ($key:expr, $($kt:ident = $kv:expr),+) => {
+        {
+            let string = t!($key);
+            $(
+                let string = string.replace(concat!("%{", stringify!($kt), "}"), &$kv.to_string());
+            )+
+            string
+        }
+    };
+    ($key:expr, $($kt:literal => $kv:expr),+) => {
+        {
+            let string = t!($key);
+            $(
+                let string = string.replace(concat!("%{", $kt, "}"), &$kv.to_string());
+            )+
+            string
+        }
+    };
+}
+
+/// Translate an item to the current locale.
+///
+/// This function is public as it is used by the `t!` macro, however is not considered part of the
+/// public interface.
+#[doc(hidden)]
+pub fn __translate(item: &str) -> &str {
+    let locale = CURRENT_LOCALE.with(Clone::clone);
+    let binding = locale.borrow();
+    translate_locale(item, binding.as_str())
+}
+
+/// Translate an item to a specific locale.
+///
+/// If the item does not exists, the key is returned. Otherwise, if item does not contain the
+/// locale, the fallback locale is used. If the fallback locale does not exist, the key is
+/// returned.
+fn translate_locale<'a>(item: &'a str, locale: &str) -> &'a str {
+    if let Some(key) = data().0.get(item) {
+        if let Some(value) = key.0.get(locale) {
+            value
+        } else if let Some(value) = key.0.get(&split_locale(locale)) {
+            value
+        } else if let Some(value) = key.0.get(FALLBACK_LOCALE) {
+            value
+        } else {
+            item
+        }
     } else {
-        set_locale_inner(FALLBACK_LOCALE);
+        item
     }
 }
 
-/// Get the current locale.
-pub fn locale() -> String {
-    rust_i18n::locale().to_string()
+/// Get the locale data.
+fn data() -> &'static Data {
+    static DATA: OnceLock<Data> = OnceLock::new();
+    DATA.get_or_init(|| {
+        toml::from_str(include_str!("../locales.toml")).expect("Failed to parse locales.toml")
+    })
 }
 
-/// Get all available locales.
-pub fn available_locales() -> Vec<&'static str> {
-    rust_i18n::available_locales!()
-}
+/// This is a map of a item name (i.e. `title_hops`, `awaiting_data`, etc.) to the locale `Item`.
+#[derive(Debug, Deserialize)]
+struct Data(HashMap<String, Item>);
 
-fn set_locale_inner(locale: &str) {
-    let all_locales = rust_i18n::available_locales!();
-    if all_locales.contains(&locale) {
-        rust_i18n::set_locale(locale);
+/// This is a map of locale keys (i.e. `en`, `zh`, etc.) to the translated value.
+#[derive(Debug, Deserialize)]
+struct Item(HashMap<String, String>);
+
+/// calculate the locale to use.
+fn calculate_locale(cfg_locale: Option<&str>, sys_locale: Option<&str>) -> String {
+    let target_locale = if let Some(locale) = cfg_locale {
+        locale
+    } else if let Some(locale) = sys_locale {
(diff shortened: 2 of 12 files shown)


nc7s:
Failure of the build-docker job seems to be a result of the way Dockerfile is written. The "caching" seems to interfere with cargo, skipping build.rs.

fujiapple852:
@nc7s I pushed a commit which fixes the Dockerfile.  I've not had time to review the rest of this yet, I will over the weekend.

fujiapple852:
aside: I think this may be broken on `master` as well, as the `locales` directory is not copied into the docker image.

fujiapple852:
@nc7s i've rebased and pushed a WIP commit that:
- replaces `phf` (and the need for the `build.rs`) with a simple `include_str`
- replaces the `RwLock` with `ArcSwap`
- Fixed an issue where the provided locale, rather than the actual locale, was displayed
- (re)added some tests

LTM what you think.

Cargo.toml
```diff
@@ -30,6 +30,7 @@ trippy-privilege = { version = "0.13.0-dev", path = "crates/trippy-privilege" }
 trippy-dns = { version = "0.13.0-dev", path = "crates/trippy-dns" }
 trippy-packet = { version = "0.13.0-dev", path = "crates/trippy-packet" }
 anyhow = "1.0.91"
+arc-swap = "1.7.1"
```
fujiapple852, line 33: Using `ArcSwap` rather than `RwLock`, but as an uncontended lock I don't image there there is an real world performance hit here. Maybe `RwLock` is better to avoid the dependency?

crates/trippy-tui/src/locale.rs
```diff
@@ -1,43 +1,125 @@
...
+        String::from(FALLBACK_LOCALE)
+    };
+    set_locale_inner(new_locale)
 }
 
-/// Get the current locale.
-pub fn locale() -> String {
-    rust_i18n::locale().to_string()
+/// Get the available locales.
+pub fn available_locales() -> Vec<&'static str> {
```
fujiapple852, line 27: Will give the sorted list of unique locales available, without assuming any particular item exists.

crates/trippy-tui/src/locale.rs
```diff
@@ -1,43 +1,125 @@
...
+}
+
+/// Translate an item to the current locale.
+///
+/// This function is public as it is used by the `t!` macro, however is not considered part of the
+/// public interface.
+#[doc(hidden)]
+pub fn __translate(item: &str) -> &str {
+    translate_locale(item, locale().load().as_str())
+}
```
fujiapple852, line 70: Separated out `translate_locale` to allow for pure/stateless testing

crates/trippy-tui/src/locale.rs
```diff
@@ -1,43 +1,125 @@
...
+/// Translate an item to a specific locale.
+///
+/// If the item does not exists, the key is returned. Otherwise, if item does not contain the
+/// locale, the fallback locale is used. If the fallback locale does not exist, the key is
+/// returned.
+fn translate_locale<'a>(item: &'a str, locale: &'_ str) -> &'a str {
+    if let Some(key) = data().0.get(item) {
+        if let Some(value) = key.0.get(locale) {
+            value
+        } else if let Some(value) = key.0.get(&split_locale(locale)) {
```
fujiapple852, line 81: reuse common split logic.  aside: we may (later..) want to introduce a "proper" `Locale` type which enforces the rules for locale strings.

crates/trippy-tui/src/locale.rs
```diff
@@ -1,43 +1,125 @@
...
+        item
+    }
+}
+
+/// Get the locale data.
+fn data() -> &'static Data {
+    static DATA: OnceLock<Data> = OnceLock::new();
+    DATA.get_or_init(|| {
+        toml::from_str(include_str!("../locales.toml")).expect("Failed to parse locales.toml")
+    })
```
fujiapple852, line 98: Avoids the need for a `build.rs`, whilst avoiding any runtime file loads.  The correctness is ensured by a unit test.

crates/trippy-tui/src/locale.rs
```diff
@@ -1,43 +1,125 @@
...
-    let all_locales = rust_i18n::available_locales!();
-    if all_locales.contains(&locale) {
-        rust_i18n::set_locale(locale);
+/// This is a map of a item name (i.e. `title_hops`, `awaiting_data`, etc.) to the locale `Item`.
+#[derive(Debug, Deserialize)]
+struct Data(HashMap<String, Item>);
+
+/// This is a map of locale keys (i.e. `en`, `zh`, etc.) to the translated value.
+#[derive(Debug, Deserialize)]
+struct Item(HashMap<String, String>);
```
fujiapple852, line 107: The key here (the locale string) could be a concrete type i.e. (`Locale`) with validation rules etc.  Overkill for now I'd say.

crates/trippy-tui/src/locale.rs
```diff
@@ -1,43 +1,125 @@
...
+
+fn set_locale_inner(locale: String) -> String {
```
fujiapple852, line 109: This logic has gone missing I think; well as doing fallback at item-translation-time there is also fallback at locale-selecton-time.

crates/trippy-tui/src/locale.rs
```diff
@@ -48,12 +130,13 @@ fn split_locale(locale: &str) -> String {
...
 }
 
-// A macro for translating a text string.
-#[macro_export]
-macro_rules! t {
-    ($($all:tt)*) => {
-        rust_i18n::t!($($all)*)
-    }
+fn locale() -> &'static ArcSwap<String> {
+    static CURRENT_LOCALE: OnceLock<ArcSwap<String>> = OnceLock::new();
```
fujiapple852, line 134: As above, `ArcSwap` may be overkill here, maybe `RwLock` is better.

Cargo.toml
nc7s: `RwLock` is just a r+w lock I happen to know ;) arc-swap is in Debian so not a problem in that regard.

crates/trippy-tui/src/locale.rs
nc7s: This can maybe be updated with the `let Some(key) = data().0.get(item) else { return item };` syntax, reducing rightward drift a bit, but otherwise it's really only aesthetic.

nc7s:
I used phf because it's more performant in this small key scenario than SipHash currently used by `std::collections::HashMap` IIRC. Did a more or less also unscientific benchmark with bench code like this:

```rust
mod locale;

use locale::*;
use rand::distributions::{Distribution, Uniform};

fn main() {
    let mut rng = rand::thread_rng();

    let all_keys: Vec<_> = data().0.keys().collect();
    let all_locales: Vec<_> = available_locales();

    let keys: Vec<_> = Uniform::new(0, all_keys.len())
        .sample_iter(&mut rng)
        .take(100)
        .map(|i| all_keys[i])
        .collect();
    let locales: Vec<_> = Uniform::new(0, all_locales.len())
        .sample_iter(&mut rng)
        .take(100)
        .map(|i| all_locales[i])
        .collect();

    for locale in locales {
        set_locale(Some(locale));
        for key in &keys {
            println!("{locale}:{key} = '{}'", t!(key));
        }
    }
}
```

and the result is ("s" means std i.e. `std::collections::HashMap`, "p" means `phf::Map`):

```
/tmp/std-phf-comp                                                                           took 5s
❯ hyperfine --warmup 50 --runs 500 s/target/release/s p/target/release/p
Benchmark 1: s/target/release/s
  Time (mean ± σ):      12.6 ms ±   0.6 ms    [User: 8.7 ms, System: 3.8 ms]
  Range (min … max):    11.5 ms …  14.7 ms    500 runs

Benchmark 2: p/target/release/p
  Time (mean ± σ):       8.5 ms ±   0.5 ms    [User: 4.9 ms, System: 3.5 ms]
  Range (min … max):     7.8 ms …  11.2 ms    500 runs

Summary
  p/target/release/p ran
    1.49 ± 0.11 times faster than s/target/release/s
```

phf is significantly faster by these numbers, but then again, probably negligible as it's on the level of nanoseconds per read.

So, ultimately your call. The code structure is kinda new to me but OK :>

Side note: I implemented "language-region to language fallback" by cutting the first two chars of `locale: &str`, assuming all language codes are ISO 639-1 i.e. two-letter. You may or may not think it's better than `.split('-').next()`.

crates/trippy-tui/src/locale.rs
nc7s: Maybe you are worried that falling back on every string selection (or item-translation-time) costs more than one time on the `set_locale()` call, and I can't deny that. Though then I'd argue locale tag splitting should also happen here, otherwise it would have the same cost burden.

Aesthetically I'm curious why this is in an "inner" function, but it's not that important.

nc7s:
> > phf is significantly faster by these numbers, but then again, probably negligible as it's on the level of nanoseconds per read.
> 
> @nc7s can `phf` be used in a similar way to `toml::from_str(include_str!("../locales.toml"))`? I'm fine to use `phf`, especially if it is faster, but would prefer to avoid the complexity of generating source code in `build.rs` for this.

I don't think so, phf is "a library to generate efficient lookup tables", **"at compile time"**. I understand build.rs is more or less ugly, but the cost of the perfect hash values might actually hurt performance if computed at runtime.

> aside: we may want a buld.rs anyway, just to have [cargo::rerun-if-changed](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed) for `crates/trippy-tui/locales.toml`.

That would ease accepting a build.rs I guess ;)

Like said earlier, though, this ultimately boils down to a difference of tens of nanoseconds per read.

> > Side note: I implemented "language-region to language fallback" by cutting the first two chars of locale: &str, assuming all language codes are ISO 639-1 i.e. two-letter. You may or may not think it's better than .split('-').next().
> 
> I'd like to swap out the adhoc string parsing for a more robust and compliant `Locale` type, similar to the ones some of the i18n crates use. Trippy should also validate the given locale string much earlier, somewhere in the `config.rs` module. For now i'll stick with the overly forgiving parsing in 0.12.0 for consistency. I've raise a follow up issue (#1432)

This smells like reinventing (some aspects of) the wheel.. though I too lean towards correctness on this matter.

fujiapple852:
@c-git the problem with using `OnceLock` in this case is the initial value is not known until runtime. Even though it will only be _set_ once at startup and not _read_ until after it is set, `__translate` is forced to specify an initial value (which it does not know) just in case it runs first.  In practise that can't actually happen, but it is a bit brittle.

crates/trippy-tui/src/locale.rs
fujiapple852: > Maybe you are worried that falling back on every string selection (or item-translation-time) costs more than one time on the set_locale() call

I think both are needed, but for different purposes. The former, at locale-selecton-time, is only really important so that the TUI shows the selected/resolved locale rather than the requested locale (so after checking if the locale exists, splitting etc).  The latter, at item-translation-time, is needed to allow partial translations (a subset of items with translations for a given language).

> Aesthetically I'm curious why this is in an "inner" function, but it's not that important.

It was just so I can add tests again the pure part of the function.  I've reworked/renamed it and added tests. Hopefully the test cases make the various locale-selecton-time fallback logic clearer.

PR #1451, opened 2024-12-23
Add CI job to enforce conventional commits

fujiapple852:
Add a CI job to enforce conventional commits.

* Add a new job `conventional-commits` to `.github/workflows/ci.yml` to enforce conventional commits using `webiny/action-conventional-commits@v1`.
* Configure the job to run on every pull requests to the `master` branch.
* Define the conventional commit types: `feat`, `fix`, `chore`, `docs`, `style`, `refactor`, `test`, `build`, and `ci`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -222,3 +222,13 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: dprint/check@v2.2
+
+  conventional-commits:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+      - name: Conventional Commits Lint
+        uses: webiny/action-conventional-commits@v1.3.0
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


c-git:
This is interesting. I have a small `commit-msg` hook that I wrote when I started trying to get accustomed to it. Copied below for reference:

```sh
#!/bin/sh

regex="^(feat|fix|build|chore|ci|docs|style|refactor|perf|test|revert)(\[.+\])?: .+"

grep -E "$regex" "$1"
if [ $? != 0 ]
then
    echo "[error] Commit message doesn't appear to in the conventional commit format"
    echo "[info] expecting \"<type>[optional scope]: <description>\""
    echo "[info] see https://www.conventionalcommits.org/en/v1.0.0/#summary for more info"
    echo "[debug] using regex: \"$regex\""
    echo "[debug] on commit message:"
    cat "$1"
    exit 1
fi
```

.github/workflows/ci.yml
```diff
@@ -221,4 +221,14 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v3
```
c-git, line 223: Maybe as you're in there you should upgrade the version of checkout in use?

PR #1452, opened 2024-12-24
ci: add job to check spelling

fujiapple852:


c-git (commented):
LGTM, the list of dictionary words looks correct as far as I can see. I was surprised to see symbols and numbers in there. That was unexpected. I checked that the words I wasn't sure of the spelling for with Google and all came back ok.

PR #1470, opened 2025-01-08
Introduce `system` address family

### LINKED ISSUE #1469: Prefer IPv6 over IPv4
We would like trippy to implement a more typical AF/source address selection policy.  Currently the default is "IPv4 then IPv6", however this is the opposite of every other typical system and application default behavior.  This leads to users or administrators who use trippy capturing data that does not show the same path/transport their application or system is using, leading to confusing diagnostics and communication.

Change the default AF/source address selection from Ipv4thenIpv6 to ipv6-then-ipv4

**Additional context**

fujiapple852:
Closes #1469


c-git:
I took a look at the code but it would take a deeper dive to answer the question I was looking for so just asking instead. Would the IP family stay consistent after startup or would it be possible for it to change per trip?

fujiapple852:
> Would the IP family stay consistent after startup or would it be possible for it to change per trip?

It will stay consistent, it is only resolved at startup.

PR #1503, opened 2025-03-13
build: exclude examples from default cargo builds

KarlEricsson:
I noticed that the hello-world and traceroute example binaries were being built by default in my nix-darwin system. This caused my system traceroute binary to be overridden by the example version.

This PR ensures these examples are only built when explicitly requested, preventing them from interfering with system binaries. If this is the intended behavior feel free to close this and I will try to solve this in nixpkgs instead.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,15 @@ members = [
   "examples/*",
 ]
 
+default-members = [
+  "crates/trippy",
+  "crates/trippy-tui",
+  "crates/trippy-core",
+  "crates/trippy-packet",
+  "crates/trippy-privilege",
+  "crates/trippy-dns",
+]
+
 [workspace.package]
 version = "0.13.0-dev"
 authors = ["FujiApple <fujiapple852@gmail.com>"]


fujiapple852:
@KarlEricsson thanks for the PR.

> This PR ensures these examples are only built when explicitly requested

The downside of this is that the examples could become broken and we would not notice. We can overcome that problem by adding `--workspace` to all commands in CI (i.e. `.github/workflows/ci.yml`) which helps, but it still has the (minor) downside that we would only discover the breakage in CI and not locally, unless we remember to pass `--workspace` to Cargo every time.

I looked at a few other options, including moving these to be "real" examples, but everything I tried has tradeoffs which are non-optional. 

Perhaps the best option is the simplest one, renaming the `traceroute` example as something like `toy-traceroute`?

> I noticed that the hello-world and traceroute example binaries were being built by default in my nix-darwin system. This caused my system traceroute binary to be overridden by the example version.

I don't know anything about `nix-darwin`, but this suggests you somehow end up with `trippy/target/release` on your path which is clearly not ideal. Are you using the trippy nix [package](https://trippy.rs/start/installation/#nix) or something custom?

KarlEricsson:
> I don't know anything about `nix-darwin`, but this suggests you somehow end up with `trippy/target/release` on your path which is clearly not ideal. Are you using the trippy nix [package](https://trippy.rs/start/installation/#nix) or something custom?

I'm using the nixpkgs package, nothing custom. As you can see [here](https://search.nixos.org/packages?channel=unstable&show=trippy&from=0&size=50&sort=relevance&type=packages&query=trippy), the the example binaries are built along with the trip binary (listed under Programs provided in the link). 

But considering the downsides you mentioned and if this is only a problem with nix packaging it's probably a better solution to solve this in nixpkgs by adding --package trippy to the cargoBuildFlags or something like that. Feel free to close this PR!

fujiapple852:
> As you can see [here](https://search.nixos.org/packages?channel=unstable&show=trippy&from=0&size=50&sort=relevance&type=packages&query=trippy), the the example binaries are built along with the trip binary (listed under Programs provided in the link).

Agreed, it is clearly not desirable to copy all three binaries. The nix package source appears to be [here](https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/tr/trippy/package.nix), would you mind raising a PR to adjust this? cc @figsoda

Separately, I think it is also worth renaming the example. I've updated this PR to do that.

fujiapple852:
I’ve merged #1508 and I see you’ve raised the PR against the nix package (thanks!) so I’ll close this PR.

PR #1525, opened 2025-03-29
Refactor windows socket addr

fujiapple852:


fujiapple852:
This ran in 40x windows simulations and did not crash.

PR #1526, opened 2025-03-29
Fix windows from length

fujiapple852:


fujiapple852:
@zarkdav FYI, I tracked down a very old bug that only started causing issues in Rust 1.85.

zarkdav:
Thank you for this fix @fujiapple852 !

zarkdav (approved):

fujiapple852:
Tested on Windows 10 VM and Windows 11 VM, no crashes.

PR #1530, opened 2025-03-31
fix(core): avoid infinite loop in is_[readable|writable]

fujiapple852:


fujiapple852:
Not needed as the code only calls `get_overlapped_result` after `wait_for_event` has returned true.

PR #1544, opened 2025-04-12
Add `Dscp` and `Ecn` columns

### LINKED ISSUE #1539: Show the payload of the ICMP time exceeded messages, this does allow to so how/where IP and L4 headers are modified along a path

I would like a mode for trippy so it does a single packet per TTL traceroute to a target and where it displays the returned payload as hex for each hop (this is the customary payload for ICMP time exceeded errors). This information is relevant for both DCSP/diffserv and ECN modifications/pathologies along a path, and having a cross platform tool to collect and report that data in an easy way would be great.

This information could also be collected e.g. by simply running a packet capture concurrent with a traceroute, or could be added to any other traceroute utility (these tend to support fewer platforms than trippy and hence are less suitable). BUt these are all pretty tedious and ill-suited to instruct normal users to collect that data.

**Additional context**

BSD's traceroute binary for IPv4 already offers the desired functionality (triggered by the `-D`switch):
IN this example `-t 224` sets the TOS bitfield to 224, which equals CS7 for the DSCP-bitfield and Not-ECT for the eCN bitfield.
```
user@123-1234567 ~ % traceroute -aID -t 224 one.one.one.one
traceroute: Warning: one.one.one.one has multiple addresses; using 1.0.0.1
traceroute to one.one.one.one (1.0.0.1), 64 hops max, 48 byte packets
 1  [AS0] 192.168.42.1 (192.168.42.1)  2.250 ms
vhtslen id  off tlprsum srcip   dstip   typ cod sum 
45e03000f6bd000001010000c0a82af90100000108000142f6bc00010000000000000000000000000000000000000000
____________________d58d________________________________________________________________________

 2  [AS6805] loopback1.0006.acln.01.ham.de.net.telefonica.de (62.52.192.120)  10.636 ms
vhtslen id  off tlprsum srcip   dstip   typ cod sum 
45e03000f6be000002010000c0a82af90100000108000141f6bc00020000000000000000000000000000000000000000
________________01__d58c________________________________

 3  [AS6805] bundle-ether17.0001.cord.01.ham.de.net.telefonica.de (62.53.11.180)  11.628 ms
vhtslen id  off tlprsum srcip   dstip   typ cod sum 
45e03000f6bf000003010000c0a82af90100000108000140f6bc00030000000000000000000000000000000000000000
________________01__d58b________________________________________________________________________0000000000000000000000000000000000000000

 4  [AS6805] ae0-0.0002.prrx.01.ham.de.net.telefonica.de (62.53.6.109)  11.141 ms
vhtslen id  off tlprsum srcip   dstip   typ cod sum 
45e03000f6c0000004010000c0a82af9010000010800013ff6bc00040000000000000000000000000000000000000000
________________01__d58a________________________________

 5  [AS0] as13335.hamburg.megaport.com (193.42.155.58)  11.602 ms
vhtslen id  off tlprsum srcip   dstip   typ cod sum 
45e03000f6c1000005010000c0a82af9010000010800013ef6bc00050000000000000000000000000000000000000000
__00____________01__d669________________________________

 6  [AS13335] one.one.one.one (1.0.0.1)  10.306 ms
vhtslen id  off tlprsum srcip   dstip   typ cod sum 
45e03000f6c2000006010000c0a82af9010000010800013df6bc00060000000000000000000000000000000000000000
000000__0000____0000____0000000000____00______3e______05
```

For each hop report the second line shows a header to give some indication about the information fields in lines 3 and 4. 
Line 3 gives the reflected header from the ICMP error payload in hexadecimal notation.
Line 4 shows `_` if the value matches the original value from the probe packet or the new value if things differ. The last hop always is special as there the TTL exceeded does not happen as the packet terminsates there

Here Hop 5 remarks the DSCP von 224 (0xe0) aka CS7 to 0 (0x0) aka CS0/DF.
The other fields that always change are the TTL (as that gets incremented compared to the first packet) and the checksum (as a consequence of the changing TTL value for the probes to the successive hops).

It would be really great for crowdsourcing measurements if sich a tool would exist for all major platforms, easy to use even for beginners and for both IPv4 and IPv6.

fujiapple852:
Closes #1539

### DIFF
--- a/crates/trippy-core/src/net/ipv4.rs
+++ b/crates/trippy-core/src/net/ipv4.rs
@@ -5,8 +5,8 @@ use crate::net::common::ErrorMapper;
 use crate::net::platform;
 use crate::net::socket::{Socket, SocketError};
 use crate::probe::{
-    Extensions, IcmpPacketCode, Probe, Response, ResponseData, ResponseSeq, ResponseSeqIcmp,
-    ResponseSeqTcp, ResponseSeqUdp,
+    Extensions, IcmpPacketCode, IcmpProtocolResponse, Probe, ProtocolResponse, Response,
+    ResponseData, TcpProtocolResponse, UdpProtocolResponse,
 };
 use crate::types::{PacketSize, PayloadPattern, Sequence, TraceId, TypeOfService};
 use crate::{Flags, Port, PrivilegeMode, Protocol};
@@ -246,7 +246,7 @@ impl Ipv4 {
         src_port: Port,
         dest_port: Port,
     ) -> Result<Option<Response>> {
-        let resp_seq = ResponseSeq::Tcp(ResponseSeqTcp::new(
+        let proto_resp = ProtocolResponse::Tcp(TcpProtocolResponse::new(
             IpAddr::V4(self.dest_addr),
             src_port.0,
             dest_port.0,
@@ -259,7 +259,7 @@ impl Ipv4 {
                 return Ok(Some(Response::TcpReply(ResponseData::new(
                     SystemTime::now(),
                     addr,
-                    resp_seq,
+                    proto_resp,
                 ))));
             }
             Some(err) => match err {
@@ -267,13 +267,13 @@ impl Ipv4 {
                     return Ok(Some(Response::TcpRefused(ResponseData::new(
                         SystemTime::now(),
                         IpAddr::V4(self.dest_addr),
-                        resp_seq,
+                        proto_resp,
                     ))));
                 }
                 SocketError::HostUnreachable => {
                     let error_addr = tcp_socket.icmp_error_info()?;
                     return Ok(Some(Response::TimeExceeded(
-                        ResponseData::new(SystemTime::now(), error_addr, resp_seq),
+                        ResponseData::new(SystemTime::now(), error_addr, proto_resp),
                         IcmpPacketCode(1),
                         None,
                     )));
@@ -306,13 +306,14 @@ impl Ipv4 {
                             (ipv4, None)
                         }
                     };
-                    self.extract_probe_resp_seq(&nested_ipv4)?.map(|resp_seq| {
-                        Response::TimeExceeded(
-                            ResponseData::new(recv, src, resp_seq),
-                            IcmpPacketCode(icmp_code.0),
-                            extension,
-                        )
-                    })
+                    self.extract_probe_proto_resp(&nested_ipv4)?
+                        .map(|proto_resp| {
+                            Response::TimeExceeded(
+                                ResponseData::new(recv, src, proto_resp),
+                                IcmpPacketCode(icmp_code.0),
+                                extension,
+                            )
+                        })
                 } else {
                     None
                 }
@@ -326,22 +327,24 @@ impl Ipv4 {
                     }
                     IcmpExtensionParseMode::Disabled => None,
                 };
-                self.extract_probe_resp_seq(&nested_ipv4)?.map(|resp_seq| {
-                    Response::DestinationUnreachable(
-                        ResponseData::new(recv, src, resp_seq),
-                        IcmpPacketCode(icmp_code.0),
-                        extension,
-                    )
-                })
+                self.extract_probe_proto_resp(&nested_ipv4)?
+                    .map(|proto_resp| {
+                        Response::DestinationUnreachable(
+                            ResponseData::new(recv, src, proto_resp),
+                            IcmpPacketCode(icmp_code.0),
+                            extension,
+                        )
+                    })
             }
             IcmpType::EchoReply => match self.protocol {
                 Protocol::Icmp => {
                     let packet = EchoReplyPacket::new_view(icmp_v4.packet())?;
                     let id = packet.get_identifier();
                     let seq = packet.get_sequence();
-                    let resp_seq = ResponseSeq::Icmp(ResponseSeqIcmp::new(id, seq, None));
+                    let proto_resp =
+                        ProtocolResponse::Icmp(IcmpProtocolResponse::new(id, seq, None));
                     Some(Response::EchoReply(
-                        ResponseData::new(recv, src, resp_seq),
+                        ResponseData::new(recv, src, proto_resp),
                         IcmpPacketCode(icmp_code.0),
                     ))
                 }
@@ -352,13 +355,13 @@ impl Ipv4 {
     }
 
     #[instrument(skip(self), level = "trace")]
-    fn extract_probe_resp_seq(&self, ipv4: &Ipv4Packet<'_>) -> Result<Option<ResponseSeq>> {
+    fn extract_probe_proto_resp(&self, ipv4: &Ipv4Packet<'_>) -> Result<Option<ProtocolResponse>> {
         Ok(match (self.protocol, ipv4.get_protocol()) {
             (Protocol::Icmp, IpProtocol::Icmp) => {
                 let echo_request = extract_echo_request(ipv4)?;
                 let identifier = echo_request.get_identifier();
                 let sequence = echo_request.get_sequence();
-                Some(ResponseSeq::Icmp(ResponseSeqIcmp::new(
+                Some(ProtocolResponse::Icmp(IcmpProtocolResponse::new(
                     identifier,
                     sequence,
                     Some(TypeOfService(ipv4.get_tos())),
@@ -369,7 +372,7 @@ impl Ipv4 {
                     extract_udp_packet(ipv4)?;
                 let expected_checksum =
                     self.calc_udp_checksum(Port(src_port), Port(dest_port), payload_length)?;
-                Some(ResponseSeq::Udp(ResponseSeqUdp::new(
+                Some(ProtocolResponse::Udp(UdpProtocolResponse::new(
                     identifier,
                     IpAddr::V4(ipv4.get_destination()),
                     src_port,
@@ -383,7 +386,7 @@ impl Ipv4 {
             }
             (Protocol::Tcp, IpProtocol::Tcp) => {
                 let (src_port, dest_port) = extract_tcp_packet(ipv4)?;
-                Some(ResponseSeq::Tcp(ResponseSeqTcp::new(
+                Some(ProtocolResponse::Tcp(TcpProtocolResponse::new(
                     IpAddr::V4(ipv4.get_destination()),
                     src_port,
                     dest_port,
@@ -1246,8 +1249,8 @@ mod tests {
         let Response::EchoReply(
             ResponseData {
                 addr,
-                resp_seq:
-                    ResponseSeq::Icmp(ResponseSeqIcmp {
+                proto_resp:
+                    ProtocolResponse::Icmp(IcmpProtocolResponse {
                         identifier,
                         sequence,
                         tos,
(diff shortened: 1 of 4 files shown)


fujiapple852:
@moeller0 would you be able to review this?

moeller0:
Sure, let's say I am willing to help, ability is work in progress. Since I am not fluent in rust I just tried to build the current version:
git pull  
 cargo install trippy --locked   

but once I run this:
sudo /Users/user/.cargo/bin/trip -m pretty --udp --tos 224 1.1.1.1

and get packet captures I see all outgoing packets marked DSCP decimal 0, aka CS0/BE/DF, so I must have screwed up something?

moeller0:
For what it worth with:
 sudo /Users/user/.cargo/bin/trip -m pretty --tcp --tos 224 1.1.1.1
I see the expected DSCP in my wireshark captures...

fujiapple852:
@moeller0 this PR doesn't actually change any functionality, it just adds datatypes for `DSCP` & `ECN` with parsing/display capabilities. It would be great if you could review the _data_ in these (i.e. the codes, names, descriptions etc). Don't worry about the Rust code too much and there is no need to run it.

-----------------

As an aside:

The `cargo install trippy --locked` command will install the latest published version of Trippy (0.12.2), _not_ the version you have checked out. 

To build the version you have checked out you can run:

```
git clone https://github.com/fujiapple852/trippy
git checkout feat-core-tos-dscp-ecn
cargo build
```

Then run it from:

```
sudo target/debug/trip -m pretty --udp --tos 224 1.1.1.1
```

Alternatively, you _can_ also install it from this PR/branch (but that isn't commonly done):

```
cargo install --git https://github.com/fujiapple852/trippy --branch feat-core-tos-dscp-ecn
```

moeller0:
>  it just adds datatypes for DSCP & ECN with parsing/display capabilities. It would be great if you could review the data in these (i.e. the codes, names, descriptions etc)

I am happy to do that later today, yesterday on first glance it looked reasonably but I did not try to scrutinise that yet.

moeller0:
As expected, when running the new code UDP works with TOS again...

crates/trippy-core/src/tos.rs
```diff
@@ -0,0 +1,206 @@
...
+    /// Not ECN-Capable Transport (00, 0 dec)
+    NotECT,
+    /// ECN Capable Transport(1) (01, 1 dec)
+    ECT1,
+    /// ECN Capable Transport(0) (10, 2 dec)
+    ECT0,
+    /// Congestion Experienced (11, 3 dec)
+    CE,
+}
+
```
moeller0, line 18: These match the data in rfc3168.

crates/trippy-core/src/tos.rs
```diff
@@ -0,0 +1,206 @@
...
+    }
+}
+
+/// Differentiated Services Code Point (`DSCP`).
+///
+/// This is used in the `DSCP` field of the `IP` header.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum DifferentiatedServicesCodePoint {
+    /// Best Effort (000 000, 0 dec)
+    BE,
```
moeller0, line 50: I know I called this best effort, but going back to rfc2474/2475 the name would be a clunky Default PHB, this was later streamlined into Default Forwarding or DF for short.
This is also one of the class-selector codepoints, specifically CS0 (rfc2474)

crates/trippy-core/src/tos.rs
```diff
@@ -0,0 +1,206 @@
...
+    /// Class Selector 3 (011 000, 24 dec)
+    CS3,
+    /// Class Selector 4 (100 000, 32 dec)
+    CS4,
+    /// Class Selector 5 (101 000, 40 dec)
+    CS5,
+    /// Class Selector 6 (110 000, 48 dec)
+    CS6,
+    /// Class Selector 7 (111 000, 56 dec)
+    CS7,
```
moeller0, line 88: The CSN definitions are all okay and match the description in rfc2474

crates/trippy-core/src/tos.rs
```diff
@@ -0,0 +1,206 @@
...
+    /// Assured Forwarding 32 (011 100, 28 dec)
+    AF32,
+    /// Assured Forwarding 33 (011 110, 30 dec)
+    AF33,
+    /// Assured Forwarding 41 (100 010, 34 dec)
+    AF41,
+    /// Assured Forwarding 42 (100 100, 36 dec)
+    AF42,
+    /// Assured Forwarding 43 (100 110, 38 dec)
+    AF43,
```
moeller0, line 74: The AFNN here match their proposed DSCH values from rfc2597.

crates/trippy-core/src/tos.rs
```diff
@@ -0,0 +1,206 @@
...
+    fn try_from(value: u8) -> Result<Self, Self::Error> {
+        match value {
+            0 => Ok(Self::NotECT),
+            1 => Ok(Self::ECT1),
+            2 => Ok(Self::ECT0),
+            3 => Ok(Self::CE),
+            _ => Err(Error::Other(format!("invalid ECN value: {value}"))),
+        }
+    }
+}
```
moeller0, line 31: Probably a silly question, why not define the decimal values in the enumerator (assuming that this is possible in rust) then the name value association would exist in one pace only?

crates/trippy-core/src/tos.rs
```diff
@@ -0,0 +1,206 @@
...
+    /// Class Selector 4 (100 000, 32 dec)
+    CS4,
+    /// Class Selector 5 (101 000, 40 dec)
+    CS5,
+    /// Class Selector 6 (110 000, 48 dec)
+    CS6,
+    /// Class Selector 7 (111 000, 56 dec)
+    CS7,
+    /// High Priority Expedited Forwarding (101 110, 46 dec)
+    EF,
```
moeller0, line 90: Matches rfc3246

crates/trippy-core/src/tos.rs
```diff
@@ -0,0 +1,206 @@
...
+    /// Other DSCP value (not defined in the standard)
+    Other(u8),
```
moeller0, line 92: I think it is worth adding a few more named PHBs/DSCPs that occasionally crop up in the wild:
VoiceAdmit/VA (rfc5865) decimal 44 (101 100)
LowerEffort/LE (rfc8622) decimal 1 (000 001)

crates/trippy-core/src/tos.rs
```diff
@@ -0,0 +1,206 @@
...
+            8 => Ok(Self::CS1),
+            16 => Ok(Self::CS2),
+            24 => Ok(Self::CS3),
+            32 => Ok(Self::CS4),
+            40 => Ok(Self::CS5),
+            48 => Ok(Self::CS6),
+            56 => Ok(Self::CS7),
+            46 => Ok(Self::EF),
+            n if n <= 0x3f => Ok(Self::Other(n)),
+            _ => Err(Error::Other(format!("invalid DSCP value: {value}"))),
```
moeller0, line 122: Not sure I read this right, but essentially all 64 possible values for the DSCP field are valid and can happen with real traffic, it is just that some have no symbolic name... that said when receiving say EF it is not guaranteed that the sending networked used that DSCP to signify the EF PHB, with DSCPs essentially anything goes. Still it is is interesting to see if/where DSCP values are changed en route, as some applications attempt to use DSCPs end to end.

moeller0 (commented):
Looks correct (I propose two additional DSCP/PHB). Since I do not know the procedure, I will submit this as comment, but I am happy to change this to any state you want.

fujiapple852:
@moeller0 i have pushed some commits to this PR which adds a (hanky) POC for two new columns showing the DSCP and ECN values from the nested Original Datagram IP header. 

You can enable them from the Trippy settings dialog to try them out.

moeller0:
Already on it, works in TUI (in that it displays something) always BE and Not-ECT in spite of the actual DSCP values as seen in wireshark for the same traffic.

Also tried:
sudo target/debug/trip -m pretty --udp --tos 224 1.1.1.1 --tui-custom-columns holsravbwdtKM
but that did not show the new columns...

moeller0:
Same result reported (BE NotECT) when I set CS7 and ECT(1) (`--tos 225`) in spite of packets actually being marked CS7*ECT(1) on the wire.

moeller0:
Probably off topic, but I like to use
sudo target/debug/trip --mode tui --icmp-extensions --dns-lookup-as-info --tui-address-mode both --tui-icmp-extension-mode all --tui-preserve-screen --dns-resolve-method cloudflare -4 one.one.one.one --tui-custom-columns holsravbwdtKM --udp --tos 225
to exercise all the nifty features that set trippy apart from its predecessors.

crates/trippy-core/src/tos.rs
fujiapple852: Added

crates/trippy-core/src/tos.rs
fujiapple852: There are some Rust proc macro crates to do this, I prefer to keep it explicit in Trippy (less "magic")

crates/trippy-core/src/tos.rs
fujiapple852: > Not sure I read this right, but essentially all 64 possible values for the DSCP field are valid and can happen with real traffic, it is just that some have no symbolic name

Exactly right. We'll allow any 6bit value and for those values which are "known" we'll map to a code (which we can display) and the rest will be shown as a number (in hex).

fujiapple852:
@moeller0 the WIP only covered ICMP tracing mode (--icmp), I've added UDP mode (--udp) and TCP (--tcp) as well. All are IPv4 only for now.

> Also tried: 
> ```
> sudo target/debug/trip -m pretty --udp --tos 224 1.1.1.1 --tui-custom-columns holsravbwdtKM
> ```

The new columns have not been added to the reports yet, only the TUI.

fujiapple852:
> Probably off topic, but I like to use
sudo target/debug/trip --mode tui --icmp-extensions --dns-lookup-as-info --tui-address-mode both --tui-icmp-extension-mode all --tui-preserve-screen --dns-resolve-method cloudflare -4 one.one.one.one --tui-custom-columns holsravbwdtKM --udp --tos 225
to exercise all the nifty features that set trippy apart from its predecessors.

Nice. I trust you've seen the [config file](https://trippy.rs/reference/configuration)?

crates/trippy-core/src/tos.rs
fujiapple852: I changed it to be `DF`.

moeller0:
> @moeller0 the WIP only covered ICMP tracing mode (`--icmp`) which would explain what you saw.
> 
> I've added UDP (`--udp`) and TCP (`--tcp`) now as well.

This is spectacular! I feel a bit daft for only trying --tcp and --udp yesterday, sorry.

> 
> Note that this is IPv4 only for now and also only covers `TimeExceeded` errors.

This already exceeds BSD traceroute by far which is the only current contender that is suitable to end users.

> 
> > Also tried:
> > ```
> > sudo target/debug/trip -m pretty --udp --tos 224 1.1.1.1 --tui-custom-columns holsravbwdtKM
> > ```
> 
> The new columns have not been added to the reports yet, only the TUI.
> 
> You might also like to add `-R dublin -P 33434` (then press `f` in the Tui) to see how the TOS is impacted by flows.

Sweet! Now I need to play with this until I find a setup that does different things t the different flow paths. ;)

moeller0:
> Nice. I trust you've seen the [config file](https://trippy.rs/reference/configuration)?

Just as general capability, I did not look into it deeply. but will do so.

moeller0:
So the feature is most excellent, having these update in TUI in real-time is really informative!
I have one question, could you either prefix numeric DSCP reports with 0x to make it clear these are hex, or convert them to decimal (the first is probably less ambiguous)?

PR #1589, opened 2025-05-18
WIP: bump `hex-literal` to 1.0.0

fujiapple852:


fujiapple852:
Blocked on edition for MSRV:

```
Caused by:
  this version of Cargo is older than the `2024` edition, and only supports `2015`, `2018`, and `2021` editions.
```

PR #1590, opened 2025-05-18
bump `etcetera` to 0.10.0

fujiapple852:


fujiapple852:
Technically this can merge now, the issue is only with the `home` v0.5.11. The `home` crate is only used by `etcetera` so leaving this for now.

`home` v0.5.11 has a higher MSRV (1.81) than we currently support. We need to keep `home` @ v0.5.9 in the `Cargo.lock` file. Unfortunately, a `cargo update` on `master` will bump `home` to the latest version.

Leaving this open until we bump to Rust 1.81.

PR #1616, opened 2025-06-20
Add sudo for convenience

dns2utf8:
Hi all

In my environment it is hard to setup sudo correctly with the paths, so I included the sudo crate to ease the use of trippy.
Let me know if the style does not match the repo.

Cheers

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2052,6 +2052,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "sudo"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


fujiapple852:
@dns2utf8 thank you for the suggestion, however I do not think this is the right fit for Trippy.

Rather than making it easier to run Trippy via sudo on Linux, I would prefer to _discourage_ using `sudo` in favour of the alternatives (i.e. `CAP_NET_RAW` or `setuid`). I should improve the documentation and examples around this to make that clearer.

Also, users of system tools like Trippy expect that it will behave in certain predictable ways in the traditions of *nix tools and I feel like they would not want it to restart itself under sudo in this way.

PR #1632, opened 2025-08-02
fix(tui): locale parsing fails for valid BCP 47 language tags (#1631)

### LINKED ISSUE #1631: locale parsing fails for valid BCP 47 language tags
Example that is not correctly handled is `zh-Hant-TW`

https://unicode.org/reports/tr35/#Unicode_language_identifier

fujiapple852:
Closes #1631


fujiapple852:
@shyuan are you able to try out this fix for the locale parsing?

shyuan:
> @shyuan are you able to try out this fix for the locale parsing?

@fujiapple852 I have tested branch https://github.com/fujiapple852/trippy/tree/fix-locale-ignore-script on macOS and the locale auto detect fallback works fine now.

Thanks.

PR #1718, opened 2025-12-23
tui: allow theming selected rows in hops table

### LINKED ISSUE #1735: allow theming selected rows in hops table

PerchunPak:
Previously, it used `hops-table-row-active-text-color` for *background* (applied it to foreground, then for some reason applied inversion filter (switch foreground and background places)). This resulted in using `bg-color` as foreground.

Now, there is just a new option for that, and there is no filter. I have changed the default value for `hops-table-row-active-text-color` (because it was technically a setting for background color, not foreground) to black (default value for `bg-color`).

Tested with #1717, but did not include it in this PR.

Closes #1735


PerchunPak:
The reason why this PR didn't work, is because `is_in_round` was always true (https://github.com/fujiapple852/trippy/pull/1717 attempted to fix it), but this line

> This is because the call to `row_highlight_style` was removed.

made me realize that I can change the modifier in that `row_highlight_style`. So this PR doesn't change anything old (unlike the previous version), just adds two new options for theming. Applying the reverse filter is just theming-unfriendly (you can't really make the colors work)

PR #1804, opened 2026-04-30
feat: add support for ipinfo lite, core, plus

### LINKED ISSUE #1784: Support for IPinfo Lite mmdb

Geoip working based on free IPinfo dataset.

It also provides ASN info, so this could be alternative for cymru lookup.

MaxMind GeoLite2 City works fine, but would be nice to have alternative.

**Additional context**

IPinfo launched free database last year [1], judging by the changes history in geoip.rs, the fact it doesn't work and [2]. I presume the lite version is not supported.

[1] https://ipinfo.io/lite
[2] https://github.com/fujiapple852/trippy/blob/01affcb37e5753e70e7d31dbdc90e654178d5733/trippy-config-sample.toml#L302-L303

ambyjkl:
Fixes #1784


crates/trippy-tui/src/geoip.rs
```diff
@@ -296,14 +352,16 @@ impl GeoIpLookup {
                 return Ok(geo.clone());
             }
             let lookup_result = reader.lookup(addr)?;
-            let city_data = if reader.metadata.database_type.starts_with("ipinfo") {
-                lookup_result
+            let city_data = match reader.metadata.database_type.strip_prefix("ipinfo ") {
+                Some("bundle_location_lite.mmdb") => lookup_result
+                    .decode::<ipinfo::IpInfoLiteGeoIp>()?
+                    .map(GeoIpCity::from),
+                Some("generic_country_free_country_asn.mmdb") => lookup_result
```
fujiapple852: @ambyjkl I'd like [confirmation](https://github.com/fujiapple852/trippy/issues/1784#issuecomment-4414863881) from IPinfo that we may freely assume the `Database Type` is a stable identifier.

If not, an alternative approach would be to match on the availability of the old and new fields and use that to determine which to use.

fujiapple852:
Thanks @ambyjkl !

crates/trippy-tui/src/geoip.rs
ambyjkl: i'm thinking while I am at it, i match all the ipinfo databases, including the paid ones:
https://ipinfo.io/developers/ipinfo-core-database
https://ipinfo.io/developers/ipinfo-plus-database

by replacing `Some("bundle_location_lite.mmdb")` with a general `Some(_)`, that would also handle any future changes on their end. `Some("generic_country_free_country_asn.mmdb")` should remain stable, since that's baked into all the existing old country-asn dbs and they are not going to make any more of those ones.

ambyjkl: The lite one and all the paid ones have the same format, it's just the more expensive, the more fields you get

fujiapple852: @ambyjkl that seems reasonable

ambyjkl: Also I just realized I broke the "IP to Geolocation Extended" db, i need to check for it specifically: `ipinfo extended_location_v2_sample.mmdb`

I think just covering until ipinfo core is enough, ipinfo lite does not have city or lat-long info and core adds that, while plus doesn't really add any field that is used by trippy

kasper93: I could swear I posted my comment few days ago, but can't find it now, so maybe I never did hit send. 

I think the old struct should be renamed to `IpInfoGeoIpLegacy` and new `IpInfoGeoIp` should support all fields, as described in their docs. Like you noticed the more expensive versions are just more fields layered on top, so it's cheap to support all of them. (while it's not cheap to use ;))

And if IpInfo breaks db schema again, without proper versioning or naming, it's on them...

ambyjkl:
Also I am seeing that instead of using mmap, the code is loading the entire mmdb into memory. I think we should use mmap on  platforms that support it. @fujiapple852 what do you think? I can add it in another PR.

crates/trippy-tui/src/geoip.rs
ambyjkl: @kasper93 I agree, I renamed the old struct to `IpInfoGeoIpLegacy` now

crates/trippy-tui/src/geoip.rs
```diff
@@ -296,14 +490,19 @@ impl GeoIpLookup {
                 return Ok(geo.clone());
             }
             let lookup_result = reader.lookup(addr)?;
-            let city_data = if reader.metadata.database_type.starts_with("ipinfo") {
-                lookup_result
+            let city_data = match reader.metadata.database_type.strip_prefix("ipinfo ") {
+                Some(
+                | "generic_country_free_country_asn.mmdb" // IP to Country + ASN Database
+                | "extended_location_v2_sample.mmdb" // IP to Geolocation Extended Database
```
fujiapple852: @ambyjkl the `_sample` here is likely wrong for the real "IP to Geolocation Extended Database" database.

crates/trippy-tui/src/geoip.rs
```diff
@@ -116,14 +116,64 @@ mod ipinfo {
...
+    /// Support the "Lite", "Core", and "Plus" database formats.
+    ///
+    /// Lite Database:
+    /// See <https://ipinfo.io/developers/ipinfo-lite-database>
+    ///
+    /// Core Database:
+    /// See <https://ipinfo.io/developers/ipinfo-core-database>
+    ///
+    /// Plus Database:
+    /// See <https://ipinfo.io/developers/ipinfo-plus-database>
```
fujiapple852, line 131: @ambyjkl I believe this should include a 4th database, namely IP to Geolocation Database which has a database_type of ipinfo standard_location_new.mmdb.

Also add to `trippy-config-sample.toml`

crates/trippy-tui/src/geoip.rs
ambyjkl: nice catch, i just pasted directly from the sample db without thinking

crates/trippy-tui/src/geoip.rs
```diff
@@ -116,14 +116,67 @@ mod ipinfo {
...
+    pub struct IpInfoGeoIp {
+        /// 42.48948
+        #[serde(default)]
+        pub latitude: Option<f64>,
+        /// -83.14465
+        #[serde(default)]
+        pub longitude: Option<f64>,
+        /// 500
+        #[serde(default)]
+        pub radius: Option<u16>,
```
fujiapple852: @ambyjkl IPinfo seems to describe this field as `INTEGER` but it does not specify what the means in MMDB terms or make any promise about the range of values (or sign!) of the field.

If we make this `i64` and then nomalize it as part of the `impl From<ipinfo::IpInfoGeoIp> for GeoIpCity`:

```
accuracy_radius: value.radius.and_then(|radius| u16::try_from(radius).ok())
```

fujiapple852:
@ambyjkl I took the liberty of pushing some changes. The main change is adding the 6x sample IPinfo `mmdb` files (which are available under the `CC-BY-4.0` license) and using these to run unit tests.

fujiapple852:
@ambyjkl one more tweak from an AI review; it now also reads the `region_code` for the new DBs which has it (Core, Plus, and IP-to-Geolocation) and maps it to `subdivision_code`.

PR #1842, opened 2026-07-08
fix(tui): centre the frequency% chart on the highest-frequency bucket

### LINKED ISSUE #90: Centre frequency% chart around the highest frequency
Currently, the frequency% barchart shows the first N round-trip-time (depending on screen width) and so can exclude the largest frequencies if they happen to be higher than can be displayed.  It would be better to centre around the highest frequency

Socialpranker:
## Summary
Fixes #90.

The frequency% barchart shows the first N round-trip-time buckets (however many fit the terminal width), sorted ascending by duration. `ratatui::widgets::BarChart` renders left-to-right and just clips whatever doesn't fit — it has no auto-centring — so if the bucket with the highest frequency happened to have a longer RTT than what fits on screen, it was pushed off-screen and never shown.

- Added `windowed_around_peak`, which slices the (already-sorted) bucket list down to however many bars fit in `rect.width` given the chart's `bar_width`/`bar_gap`, centred on the index of the highest-frequency bucket (clamped to the start/end of the data when the peak is near an edge).
- `render` now passes the windowed slice to `BarChart` instead of the full bucket list.
- `bar_width`/`bar_gap` were pulled into named constants so the windowing math and the chart itself can't drift out of sync.

## Test plan
- [x] Added 5 unit tests for `windowed_around_peak`: fits-without-windowing, peak-in-the-middle, peak-clamped-to-start, peak-clamped-to-end, width-too-small-for-one-bar.
- [x] `cargo test --package trippy-tui histogram` — 5/5 pass.
- [x] `cargo check --workspace --all-features --tests` — clean.
- [x] `cargo fmt --all -- --check` — clean.
- [x] `cargo clippy --workspace --all-features --tests -- -Dwarnings` — clean.

Note: this PR was prepared with the assistance of Claude Code, per the guidance in this repo's `AGENTS.md`. The change was validated with an actual local build/test/clippy/fmt run (via a `rust:latest` container) before submission, not just written and shipped.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/trippy-tui/src/frontend/render/histogram.rs
+++ b/crates/trippy-tui/src/frontend/render/histogram.rs
@@ -49,9 +49,17 @@ pub fn render(f: &mut Frame<'_>, app: &TuiApp, rect: Rect) {
 /// `BarChart` renders bars left-to-right and simply clips whatever doesn't fit, so without this
 /// the highest-frequency bucket can be pushed off-screen if it doesn't happen to be amongst the
 /// first buckets shown.
+///
+/// `width` is the full `rect` width passed to `render`, i.e. before the surrounding
+/// `Borders::ALL` block is applied, so two columns (one per side) are reserved for it here.
+/// The bar count itself mirrors `BarChart`'s own sizing rule (`(space + bar_gap) /
+/// (bar_width + bar_gap)`): the last visible bar doesn't need a trailing gap after it, so a
+/// plain `inner_width / bar_and_gap` would undercount by one bar whenever `inner_width mod
+/// bar_and_gap == bar_width`.
 fn windowed_around_peak(freq_data: &[(String, u64)], width: u16) -> &[(String, u64)] {
     let bar_and_gap = BAR_WIDTH + BAR_GAP;
-    let max_bars = usize::from(width / bar_and_gap).max(1);
+    let inner_width = width.saturating_sub(2);
+    let max_bars = usize::from((inner_width + BAR_GAP) / bar_and_gap).max(1);
     if freq_data.len() <= max_bars {
         return freq_data;
     }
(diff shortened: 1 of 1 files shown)


fujiapple852:
Thanks @Socialpranker. 

I trust you have tried running this locally to check how the UX feels with this change? Perhaps you can include a screenshot or screen capture video here?

Socialpranker:
> Thanks @Socialpranker.
> 
> I trust you have tried running this locally to check how the UX feels with this change? Perhaps you can include a screenshot or screen capture video here?
trippy_frequency_crop_clean

OFC! ty for fast review

fujiapple852:
@Socialpranker my robot identified two small issues with the implementation:

- `max_bars` uses the outer rectangle width, but the border removes two columns and the final bar needs no trailing gap
- The peak is selected using truncated integer percentages instead of a frequency count

Does your robot agree with my robot? Joking aside, what is your assessment?

Socialpranker:
My robot agrees with your robot on both, fixed in 04ce319.

`BarChart::render` calls `self.block.inner_if_some(area)` before doing any bar math, so the two border columns from `Borders::ALL` were never available to bars in the first place; `windowed_around_peak` was sizing against the raw rect width instead. And `BarChart`'s own bar count comes out to `(space + bar_gap) / (bar_width + bar_gap)`, which is exactly the "last bar needs no trailing gap" adjustment — a plain `width / (bar_width + bar_gap)` undercounts by one whenever the remainder equals `bar_width`. Switched to subtracting the border and using the widget's own formula, fixed the existing tests' widths (they'd been tuned to the old numbers) and added two tests that isolate each effect.

Socialpranker:
Re-checked both against the code, not just the commit message.

- (a) border + trailing gap: closed by 04ce319 — confirmed correct.
- (b) truncated-percent peak: **not actually closed by 04ce319**, despite my earlier reply. `windowed_around_peak` still picked the peak via `max_by_key` on the tuple's second field, which is `freq_pct` — the already-truncated integer percentage — `sample_frequency` never produced a separate raw count. Two buckets with different real counts can tie once truncated, and the wrong one gets picked as centre.

Fixed in d2569b5: buckets now carry `count` (raw) and `freq_pct` (display) separately; peak selection compares `count`. Added a regression test (`picks_the_peak_by_raw_count_not_truncated_percentage`) with two buckets tied at 33% (counts 100 vs 133) that fails on the old code and passes now, plus a `sample_frequency` test asserting raw counts are reported correctly.

`cargo test --package trippy-tui histogram` — 9/9 pass. `cargo clippy --package trippy-tui --all-features --tests -- -Dwarnings` and `cargo fmt --all -- --check` — clean. Rebased onto current upstream master.

On the UX capture: I didn't wire up a `TestBackend` render for this round, so I don't have a fresh screenshot/text-render for the (b) fix — happy to add one if you want it before merging.

PR #1853, opened 2026-07-17
feat(tui): add Indonesian (id) locale

fujiapple852:


crates/trippy-tui/locales.toml
```diff
@@ -1313,6 +1406,7 @@ ru = "СКО"
 es = "DesvE"
 de = "StdAbw"
 ja = "標準偏差"
+id = "Simp. Baku"
```
fujiapple852, line 1409: @s41ful maybe just "StDev" here?

crates/trippy-tui/locales.toml
```diff
@@ -1396,6 +1495,7 @@ ru = "Исх"
 es = "Sprt"
 de = "Sprt"
 ja = "送信ポート"
+id = "Port Asal"
```
fujiapple852, line 1498: @s41ful maybe just "Sprt"?

crates/trippy-tui/locales.toml
```diff
@@ -1410,6 +1510,7 @@ ru = "Назн"
 es = "Dprt"
 de = "Dprt"
 ja = "宛先ポート"
+id = "Port Tujuan"
```
fujiapple852, line 1513: @s41ful maybe just "Dprt"?

s41ful:
Yes, I agree. Keeping StDev, Sprt, and Dprt is more consistent with the other locales and better fits the TUI layout.

PR #1855, opened 2026-07-17
fix: guard packet payload slicing against malformed lengths

dkxmercury:
## Summary

Two payload accessors in `trippy-packet` slice using a length taken straight from the packet, without clamping it against the actual buffer, so a malformed packet can panic the parser.

## Root cause

**`ExtensionObjectPacket::payload`** (`icmp_extension.rs`) sliced `[minimum_packet_size()..get_length()]`. `get_length()` is a `u16` read from the object, so:

- a length below the 4 byte minimum slices `[4..n]` with `n < 4` -> `slice index starts at 4 but ends at n`
- a length beyond the buffer slices past the end -> `range end index N out of range`

`Debug` for `ExtensionObjectPacket` prints `payload()`, so even logging a malformed object hits this.

**`Ipv4Packet::payload`** (`ipv4.rs`) started at `minimum_packet_size() + ipv4_options_length(self)`, where `ipv4_options_length` is driven by the IHL nibble (`(IHL * 4).saturating_sub(20)`, up to 40). On a short packet, `start` runs past the buffer and the `[start..]` slice panics. This one is reachable from the network: the nested IPv4 header quoted inside an ICMP `TimeExceeded` / `DestinationUnreachable` error is only bounded by the outer packet, and `extract_probe_proto_resp` calls `.payload()` on it (`trippy-core/src/net/ipv4.rs`, via `extract_echo_request` / `extract_udp_packet`).

The impact is a panic (crash of the trippy process), not memory unsafety; the crate is `#![forbid(unsafe_code)]`. The same clamping is already done in `Ipv6Packet::payload`, `TcpPacket::payload` and the `*_options_raw` accessors, so these two spots look like the ones that were missed rather than a deliberate difference.

## Fix

Clamp both, matching the existing style:

```rust
// ExtensionObjectPacket::payload
let end = usize::from(self.get_length()).clamp(start, bytes.len());

// Ipv4Packet::payload
let start = std::cmp::min(minimum + ipv4_options_length(self), buf.len());
```

Output for valid packets is byte-for-byte identical; only out-of-range inputs change, from a panic to a truncated/empty slice.

## Validation

- Added four tests: extension length below minimum, extension length beyond buffer, `Debug` on a malformed extension object, and an IPv4 header claiming IHL 15 on a 20 byte packet.
- Confirmed they fail (panic) on `master` and pass with the fix.
- `cargo test -p trippy-packet` green (181 passed), `cargo clippy -p trippy-packet --all-features --tests -- -Dwarnings` clean, `cargo fmt --check` clean.
- Only `trippy-packet` touched, no behaviour change for well-formed packets.

I could only build and test the `trippy-packet` crate in my environment (`trippy-core` pulls `windows-sys`, whose `dlltool` is broken under my GNU toolchain), so the network-path reachability above is from reading `trippy-core`, not a run. CI will exercise the full workspace.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- Guard packet payload slicing against malformed lengths to prevent a panic on
+  crafted ICMP extension objects and nested IPv4
+  headers ([#1855](https://github.com/fujiapple852/trippy/pull/1855))
 - Default the `system` `address-family` to `ipv4-then-ipv6` for non-`system`
   resolvers ([#1635](https://github.com/fujiapple852/trippy/issues/1635))
 - Locale parsing fails for valid BCP 47 language tags ([#1631](https://github.com/fujiapple852/trippy/pull/1631))
--- a/crates/trippy-packet/src/icmp_extension.rs
+++ b/crates/trippy-packet/src/icmp_extension.rs
@@ -460,7 +460,13 @@ pub mod extension_object {
 
         #[must_use]
         pub fn payload(&self) -> &[u8] {
-            &self.buf.as_slice()[Self::minimum_packet_size()..usize::from(self.get_length())]
+            // `get_length()` is read straight from the packet, so a malformed
+            // object can make it smaller than the header or larger than the
+            // buffer. Clamp both ends so slicing cannot panic.
+            let bytes = self.buf.as_slice();
+            let start = Self::minimum_packet_size();
+            let end = usize::from(self.get_length()).clamp(start, bytes.len());
+            &bytes[start..end]
         }
     }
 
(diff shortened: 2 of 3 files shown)


fujiapple852:
@dkxmercury (I assume I am talking to a machine)

Silently truncating the input buffer is objectively _worse_ than explicitly crashing. When analysing network issues you need to be able to rely on your tools not to lie to you. Fail fast.

This does highlight a real bug however, namely that Trippy silently truncate input buffers in a few places today.

The question is how to handle it gracefully. 

~One idea might be to drop/skip probe responses that cannot be parsed and keep a counter of the number of such error packets. This is similar to the `Fail` column, except that is for probes which failed to send instead of probe responses that failed to parse.~ edit: except these incoming probe responses are not in themselves erroneous, they are simply larger than the static recv buffer.

dkxmercury:
What makes you think I'm a machine? The bug is real, the open question is how to fix it. A detailed, well formatted write up isn't proof that an autonomous machine wrote it. I've always written thoroughly and formatted things carefully, since long before AI.

